PR #89, opened 2023-05-24
fix: [GSFE-122] Disable the settings icon area of the home swap component

hjlee0:


2woongjae:
@hjlee0 It's already working fine, so I close it.

PR #94, opened 2023-05-24
fix: [GSFE-114] Change the font color of column names in data tables

hjlee0:


2woongjae (changes requested):
Screenshot 2023-05-26 at 6 46 11 PM
The red boxed area has a side effect.

2woongjae (approved):

PR #98, opened 2023-05-24
fix: [GSFE-115] Adjust margins of the token cards in the main menu

hjlee0:


2woongjae (changes requested):
The bottom margin still doesn't meet our requirements.

PR #99, opened 2023-05-24
fix: [GSFE-111] Anchor the header component to the top of the page

hjlee0:


2woongjae (changes requested):
Screenshot 2023-05-26 at 9 52 03 AM
These should be fixed together.

2woongjae (approved):

PR #102, opened 2023-05-24
fix: [GSFE-109] Add the token selection component

hjlee0:


2woongjae (commented):
The cursor may not be the pointer because it is unclickable. Please check.

2woongjae (approved):

PR #105, opened 2023-05-25
fix: [GSFE-107] Change the footer Component

hjlee0:


packages/web/src/components/common/footer/Footer.styles.ts
```diff
@@ -31,6 +31,8 @@ export const LeftSection = styled.div`
     color: ${({ theme }) => theme.colors.gray40};
     margin: 24px 0px;
     white-space: pre-wrap;
+    width: 100%;
+    height: 60px;
```
2woongjae, line 35: @hjlee0 Isn't it 48px?

2woongjae (commented):
Check the height value.

2woongjae (approved):

PR #112, opened 2023-05-30
fix: [GSFE-132] Fix the spacing between the header and title text

hjlee0:


jinoosss (approved):
Modified layout and component styles.

PR #113, opened 2023-05-30
fix: [GSFE-90] Change pagination component

hjlee0:


jinoosss (approved):
Added disable option to components for reusability of Swap components.

PR #115, opened 2023-05-30
fix: [GSFE-107] Change the footer Component

hjlee0:


jinoosss (approved):
Applied mobile styling and changed fixed values.

PR #132, opened 2023-07-12
feat: Add page interaction with modal

juntaepark:


packages/web/src/components/common/notification-list/NotificationList.tsx
```diff
@@ -10,11 +10,29 @@ import { TransactionGroupsType } from "@components/common/notification-button/No
...
+  modalRef: React.RefObject<HTMLDivElement | null>;
+  modalOutSideClick: () => void;
 }
 
 const NotificationList: React.FC<NotificationListProps> = ({
   txsGroupsInformation,
+  modalRef,
+  modalOutSideClick,
 }) => {
+  useEffect(() => {
```
jinoosss, line 22: Is there a reason to handle events from a Ref object passed from a NotificationList?

Being passed a Ref object creates a dependency on the parent component.

jinoosss (commented):
I have confirmed that the function works fine!

I had a few questions and left a comment.

packages/web/src/components/common/notification-list/NotificationList.tsx
```diff
@@ -10,11 +10,29 @@ import { TransactionGroupsType } from "@components/common/notification-button/No
 
 interface NotificationListProps {
   txsGroupsInformation: TransactionGroupsType[];
+  modalRef: React.RefObject<HTMLDivElement | null>;
+  modalOutSideClick: () => void;
```
jinoosss, line 14: It's a small thing, but I think function names should start with a verb.
If it's an event function, it looks good to start with on.

juntaepark: I agree. I'll fix it.

packages/web/src/components/common/notification-list/NotificationList.tsx
juntaepark: The reason why the Ref object is passed to the NotificationList is that WalletConnectorButton, the parent component, must be included in the area of ​​Ref so that event bubbling can be prevented when a button is clicked while the NotificationList is open. ! Do you have any better ideas?

jinoosss: How about handling the WalletConnectorButton event of the parent component?

juntaepark: Removed dependency on parent component and gave event capturing option as 3rd argument to modal event listener. And with stopPropagation() I blocked the event that would come after it. How about this method?

jinoosss: It's nice that it can be done on a component-by-component basis.

jinoosss (approved):
LGTM !

PR #143, opened 2023-07-25
feat: Develop TvlChart component

juntaepark:


packages/web/src/components/common/select-tab/SelectTab.tsx
```diff
@@ -1,24 +1,45 @@
...
-          key={idx}
-          className={cx({ selected: type === selectType })}
-          onClick={() => onClick(type)}
-          disabled={type === selectType}
-        >
-          {type}
-        </SelectButton>
-      ))}
+      {list.map((type, idx) =>
+        className === "chart-tab" ? (
```
jinoosss, line 23: I don't think it's explicit to handle rendering conditions by class name.
What about using explicit `type` or `enum`?

packages/web/src/components/common/select-tab/SelectTab.styles.ts
```diff
@@ -23,3 +23,18 @@ export const SelectButton = styled.button`
...
   }
 `;
+
+export const ChartSelectButton = styled.button`
+  ${mixins.flexbox("row", "center", "center")};
+  ${fonts.body10};
+  width: 60px;
+  height: 37px;
+  padding: 0px 8px;
+  border-radius: 4px, 0px, 0px, 4px;
```
jinoosss, line 33: border-radius doesn't seem to be applied.
`border-radius: 4px 0px 0px 4px;` seems to be correct.
Also, it looks like you intended for only the border-radius of the left part to be applied, I'll have to check if the same border-radius should be applied for all items.
(My guess is that the radius of the first item would be applied to the left, and the last item to the right).

packages/web/src/components/dashboard/tvl-chart-info/TvlChartInfo.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+import mixins from "@styles/mixins";
+import { fonts } from "@constants/font.constant";
+
+export const TvlChartInfoWrapper = styled.div`
+  ${mixins.flexbox("column", "flex-start", "flex-start")};
+  background-color: ${({ theme }) => theme.color.background01};
+  border-radius: 8px;
+  padding: 24px 24px 23px 24px;
+  gap: 16px;
+  width: 620px;
```
jinoosss, line 11: The chart area doesn't seem to be flexible.

packages/web/src/components/dashboard/volume-chart-info/VolumeChartInfo.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+import mixins from "@styles/mixins";
+import { fonts } from "@constants/font.constant";
+
+export const VolumeChartInfoWrapper = styled.div`
+  ${mixins.flexbox("column", "flex-start", "flex-start")};
+  background-color: ${({ theme }) => theme.color.background01};
+  border-radius: 8px;
+  padding: 24px 24px 23px 24px;
+  gap: 16px;
+  width: 620px;
```
jinoosss, line 11: The chart area doesn't seem to be flexible.

packages/web/src/containers/tvl-chart-container/TvlChartContainer.tsx
```diff
@@ -1,7 +1,67 @@
...
+    TVL_CHART_TYPE["7D"],
+  );
+
+  const { data: tvlPriceInfo } = useQuery<TvlPriceInfo, Error>({
+    queryKey: ["tvlPriceInfo"],
+    queryFn: () => fetchTvlPriceInfo(),
+    initialData: initialTvlChartPriceInfo,
+  });
+
+  const changeTvlChartType = useCallback((newType: string) => {
```
jinoosss, line 39: How do you think this code?
```
const tvlChartType = Object.values(TVL_CHART_TYPE).find(type => type === newType) || TVL_CHART_TYPE["7D"];
setTvlChartType(tvlChartType);
```

jinoosss (commented):
I left a few comments.

I'd like you to add a file for jest testing and a file for storybook testing.

packages/web/src/components/dashboard/volume-chart-info/VolumeChartInfo.styles.ts
juntaepark: I'll add. thank you.

packages/web/src/components/dashboard/tvl-chart-info/TvlChartInfo.styles.ts
juntaepark: I'll add. thank you.

packages/web/src/containers/tvl-chart-container/TvlChartContainer.tsx
juntaepark: I think the code is simpler and nicer!

packages/web/src/components/common/select-tab/SelectTab.styles.ts
juntaepark: thank you Let's check again.

packages/web/src/components/common/select-tab/SelectTab.tsx
juntaepark: Rendering by type or enum would be nice.

juntaepark:
I edited the part you commented on, please check it out.

jinoosss (approved):
LGTM !

PR #152, opened 2023-08-01
feat: Change the header

juntaepark:


packages/web/src/components/common/header/Header.styles.ts
```diff
@@ -4,39 +4,39 @@ import mixins from "@styles/mixins";
 import { Z_INDEX } from "@styles/zIndex";
 
 export const HeaderWrapper = styled.header`
-  position: fixed;
+  ${mixins.flexbox("column", "center", "center")};
   top: 0;
+  position: fixed;
   z-index: ${Z_INDEX.fixed};
+  max-width: 1920px;
```
jinoosss, line 11: If the browser size is large, this attribute will cause the header to be skewed to the left. 
Is there a reason for this attribute?

packages/web/src/constants/header.constant.ts
```diff
@@ -16,3 +16,26 @@ export const HEADER_NAV = [
     path: "/dashboard",
   },
 ];
+
+export const SIDE_MENU_NAV = [
```
jinoosss, line 20: I think the Array type is not explicit.
For example, it is hard to predict that SIDE_MENU_NAV[0] is the value for 'Governance'.
How about declaring it as an object?

packages/web/src/components/common/search-menu-modal/SearchMenuModal.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+    width: 100%;
+    .recent-searches,
+    .popular-tokens {
+      ${mixins.flexbox("row", "center", "flex-start")};
+      padding: 16px 24px 8px 24px;
+      ${fonts.body12}
+      color: ${({ theme }) => theme.color.text04};
+    }
+  }
+  li {
```
jinoosss, line 81: Missing cursor attribute.

packages/web/src/components/common/search-menu-modal/SearchMenuModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <IconSearch className="search-icon" />
+          </SearchWrapper>
+        </SearchContainer>
+        <ModalContainer>
+          <ul>
+            <div className="recent-searches">Recent Searches</div>
+            {tokens
+              .filter(item => item.searchType === "recent")
+              .map((item, idx) => (
+                <li key={idx}>
```
jinoosss, line 66: The click event for the search item is missing.

jinoosss (commented):
Nicely done.

I left a few comments.

packages/web/src/components/common/header/Header.styles.ts
juntaepark: It seems it was a mistake. Deleted. thank you !

packages/web/src/components/common/search-menu-modal/SearchMenuModal.styles.ts
juntaepark: Added. thank you :joy:

packages/web/src/constants/header.constant.ts
juntaepark: good idea. I used object.

packages/web/src/components/common/search-menu-modal/SearchMenuModal.tsx
juntaepark: Added!

jinoosss (approved):
LGTM !

PR #155, opened 2023-08-02
feat: Implement the bar graph component

jinoosss:


juntaepark:
The graph is pretty! It's very small, but what do you think about using flex-related functions? ( styles/mixins.ts - flexbox )
BarAreaGraph.styles.ts, BarGraph.styles.ts 
display: flex; flex-direction: column; -> ${mixins.flexbox("column", "flex-start", "flex-start")};

juntaepark (dismissed):
nicely done

juntaepark (approved):
nicely done

PR #174, opened 2023-08-16
feat: Refactoring the breakpoints detection logic

juntaepark:


packages/web/src/providers/gnoswap-theme-provider/GnoswapThemeProvider.tsx
```diff
@@ -1,14 +1,26 @@
...
 import { ThemeState } from "@states/index";
+import { useWindowSize } from "@hooks/common/use-window-size";
 
 const GnoswapThemeProvider: React.FC<React.PropsWithChildren> = ({
   children,
 }) => {
   const [themeKey] = useAtom(ThemeState.themeKey);
+  const { handleResize, findDeviceType } = useWindowSize();
+
+  useEffect(() => {
```
jinoosss, line 15: My guess is that you'll be registering and unregistering events frequently. 
It would be nice to improve the hooks.
Additionally, if you don't use the window size, you shouldn't save it.

jinoosss:
``` 
useEffect(() => {
    window.addEventListener("resize", handleWindowSize);
    return () => {
      window.removeEventListener("resize", handleWindowSize);
    };
  }, []);

  function handleWindowSize() {
    handleBreakpoint(window.innerWidth); 
  }
```

jinoosss (commented):
Nicely done.
It would be nice to improve the event.

packages/web/src/providers/gnoswap-theme-provider/GnoswapThemeProvider.tsx
juntaepark: good idea. The code has been cleaned up and unnecessary state has been removed.

jinoosss (approved):
Approve !

PR #178, opened 2023-08-21
Fix: Dashboard Responsive Web Develop

juntaepark:


packages/web/src/components/dashboard/volume-chart-graph/VolumeChartGraph.tsx
```diff
@@ -10,25 +11,31 @@ export interface VolumeChartGraphProps {
...
-          datas={datas}
-        />
+        <div className="graph-wrap" ref={componentRef}>
+          <BarGraph
+            className="graph"
+            width={size.width}
+            height={size.height}
+            color={theme.color.background04Hover}
+            hoverColor={theme.color.point}
+            strokeWidth={size.width * 0.022}
```
jinoosss, line 29: That ratio is something we can discuss at a later date.

packages/web/src/hooks/common/use-component-size.tsx
```diff
@@ -0,0 +1,32 @@
...
+        };
+      setSize({ width, height });
+    };
+
+    handleResize();
+
+    window.addEventListener("resize", handleResize);
+
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
```
jinoosss, line 27: It would be better if we could detect changes in the state of the `componentRef`.

jinoosss (approved):
Approve.
I left a few suggestions.

packages/web/src/components/dashboard/volume-chart-graph/VolumeChartGraph.tsx
juntaepark: i think so too.

packages/web/src/hooks/common/use-component-size.tsx
juntaepark: Edited because I thought it was a good comment.

jinoosss (approved):

PR #193, opened 2023-09-12
[GSW-15] feat: Add new item to pool list

jinoosss:


harryoh (dismissed):
👍 
Please create tickets for the tool tip on the chart.
And check the font for the text `Out of range`, `In range` on the boxes.

harryoh (approved):
👍

PR #207, opened 2023-10-06
[GSW-371] fix: Manage slippage on local storage

jinoosss:


packages/web/src/hooks/common/use-slippage.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useCallback } from "react";
+import { toDecimalNumber } from "@utils/number-utils";
+import { DEFAULT_SLIPPAGE } from "@constants/option.constant";
+
+export const useSlippage = () => {
+  const [slippage, setSlippage] = useAtom(CommonState.slippage);
+
+  const changeSlippage = useCallback(
+    (slippage: number) => {
+      let changedSlippage = slippage;
```
harryoh, line 12: How about it?
```js
const changedSlippage = Math.min(100, Math.max(0, parseInt(slippage)));
```

packages/web/src/hooks/common/use-slippage.ts
```diff
@@ -0,0 +1,29 @@
...
+      if (slippage < 0) {
+        changedSlippage = 0;
+      } else if (slippage > 100) {
+        changedSlippage = 100;
+      }
+      changedSlippage = toDecimalNumber(changedSlippage, 2);
```
harryoh, line 18: Does `slippage` need to `bignumber`?

jinoosss: No, I wanted to round to the second decimal place.
But this seems to be wrong.

harryoh (approved):
👍

PR #209, opened 2023-10-13
[GSW-416] feat: Integrate Pools API

jinoosss:


packages/web/.env.example
```diff
@@ -1 +1,5 @@
-NEXT_PUBLIC_API_URL="https://mock-api.gnoswap.io/v3/testnet"
\ No newline at end of file
+NEXT_PUBLIC_API_URL="https://dev-api.gnoswap.io/v3/testnet"
+NEXT_PUBLIC_PACKAGE_POOL_PATH="gno.land/r/pool"
+NEXT_PUBLIC_PACKAGE_POOL_ADDRESS="g1ee305k8yk0pjz443xpwtqdyep522f9g5r7d63w"
```
harryoh, line 3: It has `POOL_PATH`, but does it need to know the `address`? Is there any way to find out?

harryoh: But if the task makes more complicated, it may be okay to just use two variables.

jinoosss: We can implement an algorithm that generates an address from a path.
However, we must utilize a crypto library such as bech32.

I'm not sure which one is better, what do you think?

harryoh: First, let's use two variables.

harryoh (approved):
👍

PR #225, opened 2023-10-27
[GSW-413] fix: Wallet Improve UI

khiemldk:


packages/web/src/components/common/select-token/SelectToken.tsx
```diff
@@ -23,8 +26,10 @@ const SelectToken: React.FC<SelectTokenProps> = ({
   changeKeyword,
   changeToken,
   close,
+  callback,
 }) => {
-
+  const themeKey = useAtomValue(ThemeState.themeKey);
+  useEscCloseModal(close);
```
jinoosss, line 32: How about using this hook in a container?

packages/web/src/components/common/select-token/SelectToken.tsx
```diff
@@ -35,6 +40,7 @@ const SelectToken: React.FC<SelectTokenProps> = ({
 
   const onClickClose = useCallback(() => {
     close();
+    callback?.(true);
```
jinoosss, line 43: The callback function has already been executed in the parent container. 
Does this need to be called here as well?

packages/web/src/hooks/common/use-clear-modal.tsx
```diff
@@ -1,11 +1,12 @@
...
 import { CommonState } from "@states/index";
 import { useCallback } from "react";
+import { usePreventScroll } from "./use-prevent-scroll";
 
 export const useClearModal = () => {
-  const [, setOpenedModal] = useAtom(CommonState.openedModal);
+  const [_, setOpenedModal] = useAtom(CommonState.openedModal);
   const [, setModalContent] = useAtom(CommonState.modalContent);
-
+  usePreventScroll(_);
```
jinoosss, line 9: What does it mean?

packages/web/src/hooks/common/use-esc-close-modal.tsx
```diff
@@ -0,0 +1,19 @@
...
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener("keydown", handleEsc);
+
+    return () => {
+      window.addEventListener("keydown", handleEsc);
+    };
+  });
```
jinoosss, line 16: If you don't add `[]`, it will run multiple times. Please check.

jinoosss (commented):
@khiemldk ,
Nicely done.
I left some comments.

And I checked that you implemented a modal.
I was trying to make it work by replacing the content in a common ModalContainer, 
Do you think it's better to implement each modal separately?

khiemldk:
> @khiemldk , Nicely done. I left some comments.
> 
> And I checked that you implemented a modal. I was trying to make it work by replacing the content in a common ModalContainer, Do you think it's better to implement each modal separately?

@jinoosss , I tried to create each method separately because some previous commits separate. But maybe I can reuse the Modal container to do it better. But there are quite a few separate Modal methods being developed. What do you think if I create separate PR for this part?

packages/web/src/hooks/common/use-esc-close-modal.tsx
khiemldk: done

packages/web/src/hooks/common/use-clear-modal.tsx
khiemldk: I refactor it in the method container. This is just to prevent the content from scrolling when displaying Modal

packages/web/src/components/common/select-token/SelectToken.tsx
khiemldk: Updated

packages/web/src/components/common/select-token/SelectToken.tsx
khiemldk: updated

jinoosss:
> I tried to create each method separately because some previous commits separate. But maybe I can reuse the Modal container to do it better. But there are quite a few separate Modal methods being developed. What do you think if I create separate PR for this part?

@khiemldk, I agree.
I think it would be a good idea to create a separate PR and work on it.

jinoosss (dismissed):
👍

akstar82 (approved):

PR #226, opened 2023-10-30
[GSW-448] feat: Apply swap page query string

jinoosss:
Added URL handling based on token selection and swap direction on the swap page.


khiemldk (approved):
Okay

PR #237, opened 2023-12-01
[GSW-488] Swap Route

jinoosss:


packages/web/src/repositories/swap/swap-router-repository-impl.ts
```diff
@@ -0,0 +1,170 @@
...
+  constructor(rpcProvider: GnoProvider, walletClient: WalletClient) {
+    this.rpcProvider = rpcProvider;
+    this.walletClient = walletClient;
+  }
+
+  public findAllRoutesBy = (
+    inputToken: TokenModel,
+    outputToken: TokenModel,
+    pools: PoolModel[],
+    routeSize = 3,
```
r3v4s: can we make it variable not, hardcoded

r3v4s (approved):
2 little comments can be ignore

PR #242, opened 2023-12-07
[GSW-624] feat: use react query commonly

khiemldk:


packages/web/src/react-query/token/request.ts
```diff
@@ -0,0 +1,23 @@
+import { AxiosClient } from "@common/clients/network-client/axios-client";
+import { TokenListResponse, TokenPriceListResponse } from "@repositories/token";
+
+const API_URL = process.env.NEXT_PUBLIC_API_URL;
+
+const request = new AxiosClient(API_URL);
+
+export const getTokens = async (): Promise<TokenListResponse> => {
```
jinoosss, line 8: Fetch logic is already implemented in the token repository. 
Why did you create a new function?

packages/web/src/react-query/token/queries.ts
```diff
@@ -0,0 +1,25 @@
+import { TokenListResponse } from "@repositories/token";
+import { UseQueryOptions, useQuery } from "@tanstack/react-query";
+import { getTokenPrices, getTokens } from "./request";
+import { QUERY_KEY } from "./types";
+
+export const useGetTokensList = (
+  option?: UseQueryOptions<TokenListResponse, Error>
+) => {
+  return useQuery<TokenListResponse, Error>({
```
jinoosss, line 9: I think it would be a good idea to load the tokenRepository here.
You can use a hook called useGnoswapContext.

jinoosss (commented):
Thank you for your work.
I left a comment.

packages/web/src/react-query/token/queries.ts
khiemldk: fixed

jinoosss (approved):

PR #279, opened 2023-12-19
Gsw 677 fix calculation logic of the swap router

jinoosss:


packages/swap-router/src/swap-router/swap-router.ts
```diff
@@ -21,7 +32,63 @@ export class SwapRouter {
...
+
+    const topByDirectSwapPool = availPools
+      .filter(
+        p =>
+          [inputTokenPath, outputTokenPath].includes(p.tokenAPath) ||
+          [inputTokenPath, outputTokenPath].includes(p.tokenBPath),
+      )
+      .filter((_, index) => index < TOP_N_DIRECT_SWAPS);
+
+    const topByTVL = availPools
```
r3v4s, line 78: TVL should be aware of USD price in future

packages/swap-router/src/swap-router/swap-router.ts
```diff
@@ -2,9 +2,20 @@ import BigNumber from "bignumber.js";
...
 import { EstimatedRoute, Route, RouteWithQuote } from "./swap-router.types";
 import { makeRouteKey } from "./utility/route.util";
+import { MAX_UINT64 } from "../constants";
+
+export const TOP_N = 10;
+export const TOP_N_DIRECT_SWAPS = 2;
+export const TOP_N_TOKEN_IN_OUT = 2;
+export const TOP_N_SECOND_HOP = 1;
+export const TOP_N_WITH_EACH_BASE_TOKEN = 3;
+export const TOP_N_WITH_BASE_TOKEN = 3;
```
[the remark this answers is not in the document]
jinoosss: https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/config.ts

packages/swap-router/src/swap-router/swap-router.ts
```diff
@@ -21,7 +32,63 @@ export class SwapRouter {
...
+    const availPools = pools.filter(pool => pool.liquidity > 0);
+    const topN = 10;
+
+    const topByBaseWithTokenIn = availPools
+      .filter(
+        p => p.tokenAPath === inputTokenPath || p.tokenBPath === inputTokenPath,
+      )
+      .sort((p1, p2) => {
+        const priceOfP1 =
+          p1.tokenAPath === inputTokenPath ? p1.price : 1 / p1.price;
```
r3v4s, line 49: prefer ratio

packages/swap-router/src/swap-router/swap-router.ts
r3v4s, line 18: let's use default const from [here](https://github.com/Uniswap/smart-order-router/blob/db32bb1ba5e9cd8d525cd023934448b98d8df33d/src/routers/alpha-router/config.ts#L80-L87)

r3v4s (changes requested):
few comments left.

one need to change => change const number to same as uniswap's default chain-id's config

r3v4s (dismissed):

r3v4s (approved):

PR #283, opened 2023-12-19
[GSW-DEV] Fix: fix common page

khiemldk:


jinoosss (approved):
I think the repository should only handle raw data.

PR #302, opened 2024-01-08
[GSW-SWAP] fix: Fix Swap issue

khiemldk:


packages/web/src/containers/select-token-container/SelectTokenContainer.tsx
```diff
@@ -37,55 +37,21 @@ const customSort = (a: TokenModel, b: TokenModel) => {
...
-  } else if (a.symbol !== "GNOT" && b.symbol === "GNOT") {
-    return 1;
-  } else if (a.symbol === "GNS" && b.symbol !== "GNS") {
-    return -1;
-  } else if (a.symbol !== "GNS" && b.symbol === "GNS") {
-    return 1;
-  } else {
+
+const handleSort = (list: SortedProps[]) => {
+  const gnot = list.find(a => a.symbol === "GNOT");
```
jinoosss, line 42: How about defining symbolic values as constant values? ex) `GNOT_SYMBOL`

packages/web/src/containers/swap-container/SwapContainer.tsx
```diff
@@ -7,11 +7,28 @@ import { ThemeState } from "@states/index";
 import { useRouter } from "next/router";
 import { useSwapHandler } from "@hooks/swap/use-swap-handler";
 
+import { TokenModel } from "@models/token/token-model";
+const TOKEN_A: TokenModel = {
```
jinoosss, line 11: It seems like a good idea to modify the definition of `common/values/token-constant.ts`.

jinoosss (commented):
I left a few small comments.

packages/web/src/containers/select-token-container/SelectTokenContainer.tsx
khiemldk: done

packages/web/src/containers/swap-container/SwapContainer.tsx
khiemldk: done

dongwon8247 (approved):

PR #325, opened 2024-04-26
feat: Migrate position burn to decrease liquidity

jinoosss:
## Descriptions

- `Burn` function has disappeared from the `Position` contract.
- Change the `Burn` function to use the `DecreaseLiquidity` function.

### DIFF
--- a/packages/web/src/common/clients/wallet-client/transaction-messages/position.ts
+++ b/packages/web/src/common/clients/wallet-client/transaction-messages/position.ts
@@ -45,12 +45,48 @@ export function makePositionMintMessage(
   });
 }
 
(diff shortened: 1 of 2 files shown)


jinoosss:
@r3v4s ,
When using the DecreaseLiquidity function, 
does the token or contract need to be approved?

khiemldk (approved):

r3v4s:
> Hi @r3v4s , When using the DecreaseLiquidity function, does the token or contract need to be approved?

@jinoosss 
depends on position.
- if position doesn't have any fees to collect => no need to approve
- if position does have fees to collect, need to approve 100% amount of fee0 and fee1 to pool contract

PR #328, opened 2024-04-29
feat: Change Incentive creation time to unix time

jinoosss:
## Descriptions

When you run a function that generates an Incentive from the Stake package, 
time argument to UTC time to Unix time.

cc @r3v4s


r3v4s (approved):
thx, checked with go ide whether generated time is in seconds format (not milli seconds)

PR #329, opened 2024-04-29
fix: Fix a injection arguments type to string

jinoosss:
## Descriptions

When calling Staker's methods, change the argument type to string.

cc) @r3v4s


r3v4s (approved):
tested

PR #341, opened 2024-05-17
Refactor/UI ux enhancement

phucvarmeta:


packages/web/src/containers/gainer-and-loser-container/GainerAndLoserContainer.tsx
```diff
@@ -108,7 +108,7 @@ const GainerAndLoserContainer: React.FC = () => {
   const { isLoadingCommon } = useLoading();
 
   const gainersList = useMemo(() => {
-    return gainers.map((item: IGainer) => {
+    return (gainers ?? [])?.map((item: IGainer) => {
```
jinoosss, line 111: You can reduce the risk of out of index by reversing the order of the filter and map. 
Also, the check for `item.tokenPath === wugnotPath` could be better managed as a single variable.

packages/web/src/components/common/line-graph/LineGraph.tsx
```diff
@@ -143,11 +145,13 @@ const LineGraph: React.FC<LineGraphProps> = ({
   height = VIEWPORT_DEFAULT_HEIGHT,
   point,
   firstPointColor,
-  centerLineColor,
+  // centerLineColor,
```
jinoosss, line 148: Please remove unused line.

packages/web/src/utils/regex.ts
```diff
@@ -1 +1,3 @@
-export const REGEX_NUMBER_FORMAT = /\.00|(\.\d*?)0+$/g;
\ No newline at end of file
+export const REGEX_NUMBER_FORMAT = /\.00|(\.\d*?)0+$/g;
+
+export const cutDecimalNumberWithoutRounding = (digits: number) => new RegExp("^-?\\d+(?:\.\\d{0," + (digits || -1) + "})?");
```
jinoosss, line 3: If you want to express discarding decimals, it seems better to use `roundDown` than `cut`.

jinoosss (commented):
Great work. 
I left a couple of small comments.

Please fix the 2 test failures.
Check the toUnitFormat function in number-utils
You are calling React Query directly from DepositModal. Define QueryClientProvider in your jest file, or avoid using direct query in DepositModal.

Additionally, the ActivityInfo index column does not have a key set, please resolve the warning.

packages/web/src/utils/regex.ts
phucvarmeta: so i will change it to `roundDownDecimalNumber`

jinoosss (approved):

PR #354, opened 2024-05-25
Refactor/temp UI ux branch

phucvarmeta:


packages/web/src/hooks/swap/use-swap-handler.tsx
```diff
@@ -33,9 +33,11 @@ const findKeyByValue = (
 
 export const useSwapHandler = () => {
```
jinoosss, line 34: Could you please remove the change of use-swap-handler?
I covered snack bars and various contents in other PRs.

jinoosss (commented):
Thank you for your effort.

Remove the unused code that has been changed in the comments.
Please check the place where I didn't comment.

packages/web/src/hooks/swap/use-swap-handler.tsx
phucvarmeta: I will preserve it

jinoosss (approved):

PR #372, opened 2024-05-29
feat: Improve RPC data calls

jinoosss:
# Descriptions

This was changed because the data in ApiGetPools for RPC functions is large 
and has a negative impact on chain and interface performance.


r3v4s (approved):
confirmed 4 times call to rpc for calculation selection percent for each fee tier for same poo_path

PR #401, opened 2024-06-27
fix: migrate API field name in token prices

jinoosss:
# Descriptions

Migrate API field name in token prices. (`date` -> `time`)


r3v4s (approved):
LGTM, 👍 

One question, what's up with indent??? Is it realted to prettier?

jinoosss:
> One question, what's up with indent??? Is it realted to prettier?

Yes, the wrong prettier setting went in before.

PR #468, opened 2024-08-06
GSW-1388 Tooltip hover & scroll

AkydA:
Description
---
- don't close tooltip when hover on it
- add scroll option


jinoosss (approved):
👍

PR #503, opened 2024-08-19
GSW-764 Refactor: page-views-layout

AkydA:
GSW-764

Description
---
- Refactor page, views, layout directory structures
- `EarnAdd` component is merged into `PoolAdd`


packages/web/pages/earn/pool/stake.tsx
```diff
@@ -1,21 +1,14 @@
...
+import { DEFAULT_I18N_NS, SEOInfo } from "@constants/common.constant";
 import { SwapFeeTierInfoMap } from "@constants/option.constant";
+import useRouter from "@hooks/common/use-custom-router";
+import { useGnotToGnot } from "@hooks/token/use-gnot-wugnot";
+import PoolStake from "@views/pool/pool-stake/PoolStake";
 import { makeSwapFeeTier } from "@utils/swap-utils";
-import { DEFAULT_I18N_NS, SEOInfo } from "@constants/common.constant";
-import { serverSideTranslations } from "next-i18next/serverSideTranslations";
-import { useTranslation } from "react-i18next";
+import { useGetPoolDetailByPath } from "src/react-query/pools";
```
jinoosss, line 11: change path to `@query/*`.

packages/web/src/views/pool/pool-add/PoolAdd.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const { width } = useWindowSize();
+  const router = useCustomRouter();
+  const poolPath = router.getPoolPath();
+  const { data, isLoading } = useGetPoolDetailByPath(poolPath, {
+    enabled: !!poolPath,
+  });
+  const { getGnotPath } = useGnotToGnot();
+  const { isLoading: isLoadingCommon } = useLoading();
+
+  const hasDedicatedPool = router.asPath.includes("/pool");
```
jinoosss, line 33: This part is explicitly taken as component props.

jinoosss (approved):
Nice work !
I've left a few comments, but the behavior is fine, and it seems like a good idea to create a separate PR to work with since there are a lot of file modifications.

Also, modularization by directory would simplify the import path.

roje1 (approved):
LGTM 👍

packages/web/src/views/pool/pool-add/PoolAdd.tsx
AkydA: > This part is explicitly taken as component props.

Will do in next refactoring soon!

PR #528, opened 2024-09-09
GSW-1562 update colors

AkydA:
Description
- update colors for dark/light mode
- add cancel button
- update mock api interface
- fix some typo
- set refetch interval to 60s
- update translation


onlyhyde (approved):

packages/web/src/react-query/governance/use-get-proposals.ts
jinoosss, line 1: When new data is added, will there be duplicate data?

For example, if the first page is retrieved and new data is created before the second page is retrieved, 
it seems likely that the last data on the first page will be retrieved from the second page data.

AkydA: it could be problem, how about set caching time 0 for this query.

jinoosss (approved):

packages/web/src/react-query/governance/use-get-proposals.ts
jinoosss: It seems like a good idea to build an API based on cursors.

PR #541, opened 2024-10-15
feat: adjust request data with changes to the pool contract

jinoosss:
# Descriptions

- When creating a pool, adjust contract parameters for token ordering.
- Added caller address parameter to replace OriginCaller when minting positions.

### DIFF
--- a/packages/web/src/repositories/pool/pool-repository-impl.ts
+++ b/packages/web/src/repositories/pool/pool-repository-impl.ts
@@ -1,38 +1,15 @@
-import BigNumber from "bignumber.js";
 import { NetworkClient } from "@common/clients/network-client";
-import { PoolResponse, PoolListResponse, PoolRepository } from ".";
 import { WalletClient } from "@common/clients/wallet-client";
-import { CreatePoolRequest } from "./request/create-pool-request";
-import { TokenModel } from "@models/token/token-model";
-import {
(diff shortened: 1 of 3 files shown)


packages/web/src/repositories/pool/pool-repository-impl.ts
```diff
@@ -683,13 +684,32 @@ export class PoolRepositoryImpl implements PoolRepository {
...
+    if (!isOrdered) {
+      const reverseStartPriceSqrt =
+        startPrice !== 0 ? tickToSqrtPriceX96(priceToTick(1 / startPrice)) : 0;
+
+      return {
+        caller,
+        send: "",
+        pkg_path: POOL_PATH,
+        func: "CreatePool",
+        args: [tokenBPath, tokenAPath, fee, reverseStartPriceSqrt.toString()],
```
r3v4s, line 708: shouldn't it be `tokenAPath`, `tokenBPath` ??

jinoosss: Shouldn't the token order be reversed?

r3v4s: Oh yeah, you're right.
I was confused by variable name

onlyhyde (approved):

PR #544, opened 2024-10-26
Feature [GSW-1724]: Project-list infinite scroll

tfrg:
Apply infinite scrolling to Launchpad project list tables


packages/web/src/views/launchpad/containers/launchpad-project-list-container/LaunchpadProjectListContainer.tsx
jinoosss, line 1: In my opinion, you don't need a separate sort.
The API is already giving us sorted data, and if FE were to do a separate sort, it would result in unnatural items on infinite scroll.
https://github.com/gnoswap-labs/gnoswap-interface/blob/develop/packages/web/src/views/launchpad/containers/launchpad-project-list-container/LaunchpadProjectListContainer.tsx#L73

If you need to modify the sorted items, it would be better to do it in the API.

jinoosss (approved):

onlyhyde (approved):

PR #559, opened 2024-11-12
fix: Fix bins API lookup error for positions when page panning

jinoosss:
## Descriptions

Fixes slow page loading when navigating to the pool details page due to loading all position bins data.


tfrg (dismissed):

tfrg:
LGTM!👍

PR #587, opened 2024-12-10
feat: [GSW-1975] SwapState Loading & Insufficient Liquidity

tfrg:
## Description
- Improve handling of insufficient liquidity state in Swap page
- added the loading state to improve the UI
## Details
- Improve swap liquidity limit detection and handling in the swap feature
- Implement proper liquidity threshold tracking for better user experience
-Implement estimatedLiquidityMax-based fetch control


notJoon (approved):

packages/web/src/hooks/swap/data/use-swap.tsx
```diff
@@ -115,6 +138,8 @@ export const useSwap = ({ tokenA, tokenB, direction, slippage, swapFee = 15 }: U
   }, [direction, estimatedAmount, slippage, tokenA]);
 
   const updateSwapAmount = useCallback((amount: string) => {
+    if (!amount) return setSwapAmount(null);
+
     let newAmount = 0;
     if (!amount || BigNumber(amount).isZero()) {
```
notJoon, line 144: The `!amount` check in here is redundant since the first `if (!amount)` statement (line 141) already handles this case with an early return.

packages/web/src/hooks/swap/data/use-swap.tsx
```diff
@@ -50,7 +61,15 @@ export const useSwap = ({ tokenA, tokenB, direction, slippage, swapFee = 15 }: U
...
-      enabled: !!swapAmount && swapAmount > 0 && !!tokenA && !!tokenA.path && !!tokenB && !!tokenB.path && !isSameToken,
+      enabled:
+        shouldFetch &&
+        !!swapAmount &&
+        swapAmount > 0 &&
+        !!tokenA &&
+        !!tokenA.path &&
+        !!tokenB &&
+        !!tokenB.path &&
+        !isSameToken,
```
notJoon, line 72: Although `!!` expression is idiomatic in TypeScript, in this case it would be better to create a local variable for each logical check. This makes it clear what each condition represents and its intended purpose.

like:

```typescript
const hasValidSwapAmount = swapAmount && swapAmount > 0;
const hasValidTokenPaths = Boolean(tokenA?.path) && Boolean(tokenB?.path);
const isDifferentTokens = !isSameToken;

const isEnabled = shouldFetch && hasValidSwapAmount && hasValidTokenPaths && isDifferentTokens;
```

packages/web/src/hooks/swap/data/use-swap.tsx
```diff
@@ -20,6 +20,17 @@ export const useSwap = ({ tokenA, tokenB, direction, slippage, swapFee = 15 }: U
   const { account } = useWallet();
   const { swapRouterRepository } = useGnoswapContext();
   const [swapAmount, setSwapAmount] = useState<number | null>(null);
+  const [estimatedLiquidityMax, setEstimatedLiquidityMax] = useState<number | null>(null);
+
+  const shouldFetchData = useCallback(
+    (amount: number | null) => {
+      if (!amount) return false;
```
notJoon, line 27: This line returns `false` when `amount === 0`. If `0` is considered a valid value, it would be better to explicitly exclude only null/undefined using `amount === null` instead of `!amount`.

However, since this is related to swap functionality, it's unlikely that `amount` would ever be 0 anyway.

packages/web/src/utils/pool-utils.ts
```diff
@@ -20,7 +21,7 @@ export function makePoolPath(
   }
   const tokenAPath = tokenA.wrappedPath || tokenA.path || "";
   const tokenBPath = tokenB.wrappedPath || tokenB.path || "";
-  return [tokenAPath, tokenBPath].sort().join(":") + ":" + SwapFeeTierInfoMap[swapFeeTier].fee;
+  return [tokenAPath, tokenBPath].sort(sortTokenPaths).join(":") + ":" + SwapFeeTierInfoMap[swapFeeTier].fee;
```
notJoon, line 24: How about use `join()` consistently to concatenate all elements at one?

```suggestion
  return [...[tokenAPath, tokenBPath].sort(sortTokenPaths), SwapFeeTierInfoMap[swapFeeTier].fee].join(":");
```

notJoon (changes requested):
LGTM. left some comments

packages/web/src/hooks/swap/data/use-swap.tsx
tfrg: @notJoon 
Currently, the `!amount` check is working exactly as intended.
API calls should also not be made when amount is zero, which includes both of the following situations:
1. If 0 is entered as the final value
2. If a zero is temporarily entered while entering a decimal value (for example, ‘0’ → ‘0.’ → ‘0.1’ → ‘0.12’)

Therefore, the current implementation of checking for both null and zero via !amount is appropriate.

packages/web/src/hooks/swap/data/use-swap.tsx
tfrg: Removed unnecessary conditional checks.

packages/web/src/hooks/swap/data/use-swap.tsx
tfrg: For clarity of purpose, we managed the logic checks as local variables.

packages/web/src/utils/pool-utils.ts
tfrg: We've modified the code to make it consistent.

notJoon (dismissed):
👍

onlyhyde (approved):

notJoon (approved):

PR #596, opened 2024-12-23
[GSW-1987] optimize swap route

tfrg:
## Description
- Optimised the processing of swap data in FE
## Details
- Unnecessarily many retries
- Removes forced latency of 0.3ms for data processing

### DIFF
--- a/packages/web/src/components/swap/confirm-swap-modal/ConfirmSwapModal.tsx
+++ b/packages/web/src/components/swap/confirm-swap-modal/ConfirmSwapModal.tsx
@@ -14,8 +14,8 @@ import { PriceImpactStatus } from "@hooks/swap/data/use-swap-handler";
 import { SwapResultInfo } from "@models/swap/swap-result-info";
 import { swapDirectionToGuaranteedType, SwapSummaryInfo } from "@models/swap/swap-summary-info";
 import { SwapTokenInfo } from "@models/swap/swap-token-info";
-import { formatOtherPrice } from "@utils/new-number-utils";
-import { toNumberFormat } from "@utils/number-utils";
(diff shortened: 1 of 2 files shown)


dongwon8247 (approved):

packages/web/src/components/swap/confirm-swap-modal/ConfirmSwapModal.tsx
```diff
@@ -123,6 +127,22 @@ const ConfirmSwapModal: React.FC<ConfirmSwapModalProps> = ({
...
+      if (!tokenBUSD || tokenBUSD === 0) return "-";
+      return `($${convertToKMB(floorNumber((tokenBUSD / Number(tokenBAmount)) * swapRate).toFixed(3), {
+        isIgnoreKFormat: true,
+      })})`;
+    } else {
+      if (!tokenAUSD || tokenAUSD === 0) return "-";
+      return `($${convertToKMB(floorNumber((tokenAUSD / Number(tokenAAmount)) * swapRate).toFixed(3), {
+        isIgnoreKFormat: true,
+      })})`;
+    }
```
notJoon, line 143: seems duplicate. how about make this as an function?

```typescript
const formatPrice = (usd: number, amount: number): string => {
      if (!usd || !amount) return '-'; // xxx: not sure about the condition.

      const calculated = (usd / amount) * swapRate;
      const flooredValue = floorNumber(calculated.toFixed(3), {
        isIgnoreKFormat: true,
      });
      return `($${convertToKMB(flooredValue, { isIgnoreKFormat: true })})`;
    };

    if (swapRateAction === 'ATOB') {
      return formatPrice(tokenBUSD, tokenBAmount);
    } else {
      return formatPrice(tokenAUSD, tokenAAmount);
    }
```

PR #598, opened 2024-12-29
[GSW-2033] swap renewal

tfrg:
## Description
- Renewal of the Swap info and (full) quantity input fields on the Swap page
## Details
- Add Token price chart
- Add Token Pair transaction list
- Add Max button UI to quantity input field


packages/web/src/components/common/header/search-menu-modal/SearchMenuModal.tsx
```diff
@@ -77,23 +78,39 @@ const SearchMenuModal: React.FC<SearchMenuModalProps> = ({
...
   const tokenNamePopularRef = useRef(popularTokens.map(() => React.createRef<HTMLSpanElement>()));
   const tokenNameRecentsRef = useRef(recents.map(() => React.createRef<HTMLSpanElement>()));
   const recentPriceRef = useRef(recents.map(() => React.createRef<HTMLDivElement>()));
   const popularPriceRef = useRef(popularTokens.map(() => React.createRef<HTMLDivElement>()));
 
+  const [widthListPopular, setWidthListPopular] = useState<number[]>(popularTokens.map(() => 0));
+  const [widthListRecent, setWidthListRecent] = useState<number[]>(recents.map(() => 0));
+  const [tokenNameRecentWidthList, setTokenNameRecentWidthList] = useState<number[]>(recents.map(() => 0));
+  const [tokenNamePopularWidthList, setTokenNamePopularWidthList] = useState<number[]>(popularTokens.map(() => 0));
+
```
notJoon, line 93: It seems possible to create a hook that returns `number[]` and `setWidthList` using generics.

something like:

```typescript
function useWidthList<T>(items: T[]): [number[], React.Dispatch<React.SetStateAction<number[]>>] {
  return useState<number[]>(items.map(() => 0));
}
```

packages/web/src/components/common/line-graph/LineGraph.tsx
```diff
@@ -571,6 +587,25 @@ const LineGraph: React.FC<LineGraphProps> = ({
     return parseTimeTVL(datas[currentPointIndex]?.time);
   }, [currentPointIndex, datas, displayLastDayAsNow]);
 
+  useEffect(() => {
+    if (currentPointIndex >= 0) {
+      const currentDate =
+        displayLastDayAsNow && datas.length - 1 === currentPointIndex
+          ? parseTimeTVL(getLocalizeTime(new Date().toString()))
```
notJoon, line 594: If you perform index range checking first for early return, it seems like it would improve readability. 

Actually the current version is fine as well.

packages/web/src/hooks/swap/data/use-swap-handler.tsx
```diff
@@ -104,6 +104,14 @@ function compareAmountFn(amountA: string | number | bigint, amountB: string | nu
 function handleAmount(changed: string, token: TokenModel | null) {
   let value = changed;
   const decimals = token?.decimals || 0;
+
+  // Check if input exceeds decimal places
+  const parts = changed.split(".");
+  if (parts.length > 1 && parts[1].length > decimals) {
+    // Signal invalid input
+    return { isValid: false, value: changed };
+  }
```
notJoon, line 113: This can be simplify, if using optional chain here.

```suggestion
  if (changed.split('.')[1]?.length > decimals) {
    return { isValid: false, value: changed };
  }
```

packages/web/src/layouts/swap/components/swap-card/swap-card-content/SwapCardContent.tsx
```diff
@@ -121,6 +121,10 @@ const SwapCardContent: React.FC<ContentProps> = ({
     return swapTokenInfo.tokenBAmount;
   }, [swapTokenInfo.tokenBAmount, tokenB?.decimals]);
 
+  const hasTokenABalance = useMemo(() => {
+    return swapTokenInfo.tokenABalance !== "-" && swapTokenInfo.tokenABalance !== "0";
+  }, [swapTokenInfo.tokenABalance]);
```
notJoon, line 126: In this case, you can also use `include` method.

```suggestion
    const hasTokenABalance = useMemo(
  () => !["-", "0"].includes(swapTokenInfo.tokenABalance),
  [swapTokenInfo.tokenABalance]);
```

packages/web/src/layouts/swap/components/swap-card/swap-card-content/SwapCardContent.tsx
```diff
@@ -164,7 +177,11 @@ const SwapCardContent: React.FC<ContentProps> = ({
         <div className="amount-container">
           <input
             id={tokenB?.priceID}
-            className={`amount-text ${isLoading && direction === "EXACT_IN" ? "text-opacity" : ""}`}
+            className={`amount-text ${
+              (isLoading && direction === "EXACT_IN") || (isRefetching && direction === "EXACT_IN")
```
notJoon, line 181: `(A && C) || (B && C)` is equivalent to `(A || B) && C`. Therefore, you can simplify this condition like this:

```suggestion
              (isLoading || isRefetching) && direction === "EXACT_IN"
```

packages/web/src/layouts/token-detail/components/token-swap/TokenSwap.tsx
```diff
@@ -108,12 +109,9 @@ const TokenSwap: React.FC<TokenSwapProps> = ({
...
   }, [changeTokenAAmount, connected, dataTokenInfo]);
 
-  const handleAutoFillTokenB = useCallback(() => {
-    if (connected) {
-      const formatValue = parseFloat(dataTokenInfo.tokenBBalance.replace(/,/g, "")).toString();
-      changeTokenBAmount(formatValue);
-    }
-  }, [changeTokenBAmount, connected, dataTokenInfo]);
+  const hasTokenABalance = useMemo(() => {
+    return swapTokenInfo.tokenABalance !== "-" && swapTokenInfo.tokenABalance !== "0";
```
notJoon, line 113: same as above, you can use `include` function here

notJoon (changes requested):
I've added a few improvements. But LGTM

packages/web/src/components/common/header/search-menu-modal/SearchMenuModal.tsx
tfrg: Yes, as you said, I already created a `useElementWidthList` hook for managing the width state, 
and the width measurement logic is also redundant, so I encapsulated that as a hook as well.

The `useEffects` of the current component:
```typescript
useEffect(() => {
  const widthValues: number[] = [];
  popularPriceRef.current.forEach(ref => {
    if (ref.current) {
      const width = ref.current.getBoundingClientRect().width;
      widthValues.push(width);
    }
  });
  setWidthListPopular(widthValues);
}, [popularPriceRef]);

// ... three more useEffects with similar patterns
```

packages/web/src/components/common/line-graph/LineGraph.tsx
tfrg: Yes, I think the code will be more readable if you modify it with an early return pattern.

The current code is clean enough, but I think the following changes would make the code flow clearer:

```typescript
useEffect(() => {
  if (currentPointIndex < 0) {
    onLineGraphMouseMove?.(undefined, undefined);
    return;
  }

  const currentDate =
    displayLastDayAsNow && datas.length - 1 === currentPointIndex
      ? parseTimeTVL(getLocalizeTime(new Date().toString()))
      : parseTimeTVL(datas[currentPointIndex]?.time);

  const formattedValue = popupYValueFormatter
    ? popupYValueFormatter(datas[currentPointIndex]?.value)
    : formatPrice(datas[currentPointIndex]?.value);

  onLineGraphMouseMove?.(datas[currentPointIndex], {
    ...currentDate,
    value: formattedValue,
  });
}, [currentPointIndex, datas, displayLastDayAsNow, popupYValueFormatter]);

- Handle exception cases first and return early
- Reduce nesting of main logic
- Ensures exactly the same behavior as existing code

File: packages/web/src/hooks/swap/data/use-swap-handler.tsx (hunk 3 above)
Comment (tfrg) [reply packages/web/src/hooks/swap/data/use-swap-handler.tsx:113]:
```typescript
// Before
if (changed.split('.')[1]?.length > decimals) {
  return { isValid: false, value: changed };
}

// After
if (changed.includes('.') && changed.split('.')[1].length > decimals) {
  return { isValid: false, value: changed };
}
```

If check for decimal points first with `includes('.')` :
- More intuitive representation of decimal point presence
- Avoids unnecessary `split()` operations if no decimal point is present

Thank you for your suggestion:
- Optimize memory by removing unnecessary variable assignments
- Leaner, more intuitive code while maintaining the same behavior

packages/web/src/layouts/swap/components/swap-card/swap-card-content/SwapCardContent.tsx
tfrg: Currently, we only have 2 comparisons, so I think the existing method is clearer and more efficient. 
But thanks for the suggestion, I think the includes method you suggested will be more maintainable if we have more than 3 values to compare in the future.

packages/web/src/layouts/swap/components/swap-card/swap-card-content/SwapCardContent.tsx
tfrg: Thanks, that's a great tip.

packages/web/src/layouts/token-detail/components/token-swap/TokenSwap.tsx
tfrg: Yes, you're right, I think the current code is clearer and more efficient since we only have two values to compare, “-” and “0”.

However, if in the future we have more than three values to compare (e.g. “-”, “0”, “undefined”, etc.), then I think it would be cleaner and more maintainable to use includes as shown below:

notJoon (approved):
LGTM 👍

jinoosss (approved):

PR #602, opened 2025-01-09
[GSW-2048] Swap Issue 2

tfrg:
## Description
- Fixed data not being called when selecting a token after entering a swap quantity.
- Execute Wrap/Unwrap

### DIFF
--- a/packages/web/src/hooks/swap/data/use-swap-handler.tsx
+++ b/packages/web/src/hooks/swap/data/use-swap-handler.tsx
@@ -1102,12 +1102,12 @@ export const useSwapHandler = () => {
 
     // Handle default values when no user input exists
     if (estimateFlagRef.current === 0) {
-      if (!!defaultTokenAAmount) {
+      if (defaultTokenAAmount) {
         estimateFlagRef.current += 1;
         changeTokenAAmount(defaultTokenAAmount);
         return;
       }
-      if (!!defaultTokenBAmount) {
+      if (defaultTokenBAmount) {
         estimateFlagRef.current += 1;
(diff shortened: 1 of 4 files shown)


packages/web/src/hooks/swap/data/use-swap-handler.tsx
```diff
@@ -1095,6 +1094,51 @@ export const useSwapHandler = () => {
...
+      return;
+    }
+
+    // Check for user input only for non-same token cases
+    if (!isSameToken) {
+      const isUserInput = tokenAAmount !== "" || tokenBAmount !== "";
+      if (isUserInput) {
+        return;
+      }
+    }
```
notJoon, line 1129: consider reordering conditions. cheking `!isSameToken` first allows removing redundant `isSameToken` checks in another conditions.

```suggestion
    // Check for user input only for non-same token cases
    if (!isSameToken) {
      const isUserInput = tokenAAmount !== "" || tokenBAmount !== "";
      if (isUserInput) {
        return;
      }
    }

    // Special handling for isSameToken case
    if (tokenAAmount && !tokenBAmount) {
      setTokenBAmount(tokenAAmount);
      return;
    }
```

notJoon (changes requested):

packages/web/src/hooks/swap/data/use-swap-handler.tsx
tfrg: Code is now more readable, taking into account condition reordering!

notJoon (approved):

dongwon8247 (approved):

PR #645, opened 2025-03-27
[GSW-2177] Pool Data Handling

tfrg:
## Description
This is the PR that the frontend responded to after changing the Pool data structure.

## Details
- Remove /incentivize/pools
- Unify with /pools
- FeeTier Logic
- CurrentTick Graph
- RewardToken Logic


packages/web/src/components/common/pool-graph/pool-graph-svg/PoolGraphSVG.tsx
```diff
@@ -86,6 +86,8 @@ const PoolGraphSVG = forwardRef<SVGSVGElement, PoolGraphSVGProps>(
     const { defaultMinX, minX, maxX } = d3Position;
     const centerX = currentTick ?? (minX && maxX ? (minX + maxX) / 2 : 0);
 
+    const hasCurrentTick = React.useMemo(() => currentTick !== null && currentTick !== undefined, [currentTick]);
```
notJoon, line 89: niptick: `!= null` checks that the value is neither null nor undefined

```suggestion
    const hasCurrentTick = React.useMemo(() => currentTick != null, [currentTick]);
```

packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
```diff
@@ -123,10 +129,36 @@ const RewardTooltipContent: React.FC<RewardTooltipContentProps> = ({ rewardInfo
...
+        return null;
+      }
+
+      if (accum === null) {
+        return current.usd;
+      }
+
+      if (current.usd === null) {
+        return accum;
+      }
```
notJoon, line 148: This condition is already filtered in the first conditional statement. It seems to be dead code.

packages/web/src/utils/reward-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @returns Reward types for display
+ */
+export const mapToDisplayRewardType = (rewardType: RewardType): DisplayRewardType => {
+  if (rewardType === "SWAP_FEE") return "SWAP_FEE";
+  if (rewardType === "EXTERNAL_REWARD") return "EXTERNAL_REWARD";
+  if (rewardType === "INTERNAL_TIER_1" || rewardType === "INTERNAL_TIER_2" || rewardType === "INTERNAL_TIER_3") {
+    return "INTERNAL_REWARD";
+  }
+  return "NONE";
+};
```
notJoon, line 15: How about using a switch statement here?

Additionally, the final `return "NONE"` is unreachable code. Since `RewardType` is defined as a union type, the value passed to the function must always be one of the defined types.

notJoon (dismissed):

notJoon (approved):
👍

PR #651, opened 2025-04-01
Fix: TokenPrice Chart decimal

tfrg:
## Description
A PR that handles known issues.

## Details
- Handling when latestPrice data is not available
- SwapChart decimals not applied

### DIFF
--- a/packages/web/src/layouts/swap/components/swap-info-chart/swap-token-info/SwapTokenHeader.tsx
+++ b/packages/web/src/layouts/swap/components/swap-info-chart/swap-token-info/SwapTokenHeader.tsx
@@ -48,7 +48,7 @@ const SwapTokenHeader = ({ tokenInfo, currentPrice, chartData, containerWidth }:
 
   const displayPrice = React.useMemo(() => {
     const price = nullish.handleFalsy(chartData?.value, currentPrice);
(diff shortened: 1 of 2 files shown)


packages/web/src/layouts/swap/components/swap-info-chart/swap-token-info/SwapTokenHeader.tsx
```diff
@@ -48,7 +48,7 @@ const SwapTokenHeader = ({ tokenInfo, currentPrice, chartData, containerWidth }:
 
   const displayPrice = React.useMemo(() => {
     const price = nullish.handleFalsy(chartData?.value, currentPrice);
-    return `${formatPrice(price, { lessThan1Significant: 2 })}`;
+    return `${formatPrice(price, { isKMB: false, lessThan1Significant: 2, forcedDecimals: true })}`;
```
notJoon, line 51: Just a question: does `isKMB` always have to be set manually?

notJoon (approved):
LGTM

packages/web/src/layouts/swap/components/swap-info-chart/swap-token-info/SwapTokenHeader.tsx
tfrg: It doesn't always have to be set manually. The default value is true.

PR #654, opened 2025-04-03
[GSW-2199, GSW-2204] Implement Leaderboard feature

tfrg:
## Description
This PR implements the leaderboard feature that displays user rankings and information.

### API Integration
- Implemented LeaderboardRepository interface with methods for fetching leaderboard data
- Added getLeaderboard method to retrieve all leaderboard users
- Added getLeaderboardByAddress method to retrieve specific user information
- Fixed type issues in API response handling
### UI Components
- Enhanced UserColumn component with better code structure and responsiveness
- Implemented pagination for the leaderboard table
- Added support for mobile, tablet, and web layouts
### Optimistic UI
- Implementing immediate UI updates before API response when user invisibility status changes
- Enforced optimistic update pattern for toggle hidden status operations
- Added error handling logic to roll back to original state when API fails
- Optimised loading state management for better user experience
### Data Processing
- Added keyword filtering functionality for searching users
- Implemented pagination logic with configurable items per page
- Added user's own information display when connected wallet is available
### Testing & Development
- Added mock implementation for LeaderboardRepository for testing and development
- Generated realistic random data for testing various scenarios


packages/web/src/layouts/leaderboard-layout/components/point-composition/PointComposition.tsx
```diff
@@ -1,21 +1,55 @@
...
+
+  const displayXGNSPoint = React.useMemo(() => {
+    if (!governancePoint) return "-";
+    return `${numberToFormat(governancePoint)}`;
+  }, [governancePoint]);
+
+  const displayReferralPoint = React.useMemo(() => {
+    if (!referralPoint) return "-";
+    return `${numberToFormat(referralPoint)}`;
+  }, [referralPoint]);
```
notJoon, line 51: There is duplication in the point format conversion logic. It would be good to extract this into a separate function. Also, by integrating all `useMemo` into one, you can calculate all point values at once.

Structuring the data makes it easier to add or delete items later, so it's worth considering defining tooltip point items as a `pointItems` array.

Additionally, declaring an interface for props can ensure type safety.

For example, you could do it like this (pseudo-code level):

```ts
interface PointCompositionProps {
  totalPoint: string;
  swapPoint: string;
  positionPoint: string;
  stakingPoint: string;
  governancePoint: string;
  referralPoint: string;
  isMobile: boolean;
}

// Extract formatting logic to a reusable function
const formatPoint = (point: string): string => {
  if (!point) return "-";
  return numberToFormat(point);
};

const PointComposition = ({
  // ...
}: PointCompositionProps) => {
  const displayPoints = React.useMemo(() => ({
    total: formatPoint(totalPoint),
    swap: formatPoint(swapPoint),
    position: formatPoint(positionPoint),
    staking: formatPoint(stakingPoint),
    governance: formatPoint(governancePoint),
    referral: formatPoint(referralPoint),
  }), [totalPoint, swapPoint, positionPoint, stakingPoint, governancePoint, referralPoint]);

  // rendering point items
  const pointItems = [
    { label: "Swap", value: displayPoints.swap },
    { label: "Position", value: displayPoints.position },
    { label: "Position Staking", value: displayPoints.staking },
    { label: "xGNS Staking", value: displayPoints.governance },
    { label: "Referral", value: displayPoints.referral },
  ];

  // ...
  // update return accordingly. for example:
  // <Flex key={item.label}>
  //             <Label>{item.label}</Label>
  //             <FontWeight500>{item.value}</FontWeight500>
  //  </Flex>
}
```

packages/web/src/repositories/leaderboard/leaderboard-repository-impl.ts
```diff
@@ -0,0 +1,96 @@
...
+    const response = await this.networkClient.get<{ data: GetLeaderboardResponse }>({
+      url: `/leaderboard?${queries.filter(query => !!query).join("&")}`,
+    });
+
+    if (!response?.data?.data) {
+      return nullLeaderboardInfo;
+    }
+
+    const data: GetLeaderboardResponse = response.data.data;
+    return data;
```
notJoon, line 39: In this case, the type inference is automatic, so there's no problem with returning directly. If you apply this fix, you'll need to apply it to other functions as well.
```suggestion
    return response.data.data;
```

packages/web/src/repositories/leaderboard/leaderboard-repository-impl.ts
```diff
@@ -0,0 +1,96 @@
...
+  nullLeaderboardInfo,
+  UpdateLeaderboardHiddenStateResponse,
+} from "./response";
+
+export class LeaderboardRepositoryImpl implements LeaderboardRepository {
+  private networkClient: NetworkClient | null;
+
+  constructor(networkClient: NetworkClient | null) {
+    this.networkClient = networkClient;
+  }
```
notJoon, line 17: Since the constructor accepts a NetworkClient | null type, networkClient can be null. However, the current methods in the class check for if (!this.networkClient), which suggests that a client is required for all operations.

```suggestion
  constructor(private readonly networkClient: NetworkClient) {}
```

Therefore, if the constructor requires a non-null client, it seems possible to remove all these checks.

notJoon (changes requested):
I've left some suggestions, please check them out. thanks

packages/web/src/layouts/leaderboard-layout/components/point-composition/PointComposition.tsx
tfrg: I hadn't thought about de-duplicating code, thanks.

packages/web/src/repositories/leaderboard/leaderboard-repository-impl.ts
tfrg: Thank you for the refactoring suggestion. I'm going to review it and keep the current approach.

keep it this way for the following reasons:
1. Value of explicit type declaration: By explicitly declaring the GetLeaderboardResponse type, the structure and intent of the return data is immediately clear to anyone reading the code.

2. Codebase consistency: As you mentioned, this change would require modifying all of our other functions, and we feel that making extensive changes for a small optimisation is a low priority at this point.

3. Ease of debugging: Having an intermediate variable makes it easier to inspect the value when debugging, which helps with troubleshooting.

packages/web/src/repositories/leaderboard/leaderboard-repository-impl.ts
tfrg: I've reviewed @notJoon's suggestion, but I think it's better to keep the current approach (allowing NetworkClient | null), for the following reasons

Maintaining the flexibility of the dependency injection pattern
‘Since we're using the pattern of injecting clients, I think it's better to keep it the way it is.’

1. The value of explicit type declarations: By explicitly declaring the type NetworkClient | null, we make it clear to anyone reading the code that this class can be initialised without a client.

2. Codebase consistency: It's likely that other repository classes currently follow a similar pattern. Changing just one class can break the consistency of your codebase.

3. Flexible instance creation: Allowing nulls gives you the flexibility to respond to different scenarios (lazy initialisation, conditional client setup, etc.).

4. Ease of testing: In a test environment, it's easy to create and test instances without real network clients.

notJoon: Here's my thoughts:

* **Maintaining Flexibility in Dependency Injection**: While allowing `null` in dependency injection patterns may offer more flexibility, this approach can undermine the intent of dependency injection. One of the core principles of dependency injection is managing component coupling through clear contracts. For essential dependencies, requiring non-null values in the constructor better aligns with this principle. [^1]

* **Value of Explicit Type Declarations**: Explicitly declaring types as `NetworkClient | null` may help with code readability, but if you need to perform `null` checks in every method, this implies that the dependency should never actually be `null`. In this case, filtering to only accept the `NetworkClient` type could be considered more explicit.

* **Codebase Consistency**: If other classes in the codebase follow similar patterns, maintaining consistency is important (as doing things differently could cause problems). However, I believe gradually improving the entire codebase with better patterns would be beneficial in the long term.

* **Flexible Instance Creation**: While allowing `null` provides flexibility in scenarios like lazy initialization or conditional client setup, these cases can be adequately addressed through other approaches. Of course, this point is somewhat ambiguous since implementing such changes might require significant effort.

* **Testability**: I understand that in test environments, it's easier to create and test instances without actual network clients. However, using mocks or stubs to replace real dependencies is a common testing approach.

I fully understand the reasoning for maintaining the current code structure. However, in the long term, I believe improving the code to provide clear contracts and type safety would be better. i'll approve this PR. But please consider this in the future work. Thank you.

[^1]: https://en.wikipedia.org/wiki/Dependency_inversion_principle

tfrg: thank you for your detailed review. Your deep understanding of the dependency injection pattern and your thoughts on improving code quality were very helpful.

As you say, I think incrementally improving the codebase is a better way to go in the long run

For now, we'll keep the PR as is, but we'll prioritise these improvements in future sprints. Thanks again, for your insightful suggestions on improving code quality.

notJoon (approved):

PR #657, opened 2025-04-08
[GSW-2204] Referral Address closure issue

tfrg:
## Description
This PR resolves an issue where the latest referralAddress value was not being properly reflected in various functions due to a closure issue. The functions were capturing the initial referralAddress value when they were defined, but were not updating when this value changed.

## What's changed
- Added a referralAddressRef using useRef to the useReferral hook so that we can always keep track of the latest referralAddress value.
- Added an effect to update referralAddressRef.current whenever the referralAddress changes.
- Added a getCurrentReferralAddress method to the return value of the useReferral hook that always returns the latest referral address value.
- This ensures that even if a component captures a function early in its lifecycle, calling that function will always return the latest value.
- We've pushed this improvement across the board so that it applies wherever referralAddress is used, not just in swap functions.

### DIFF
--- a/packages/web/src/hooks/common/use-referral.tsx
+++ b/packages/web/src/hooks/common/use-referral.tsx
@@ -25,7 +25,7 @@ export const useReferral = () => {
   const [referralAddress, setReferralAddress] = React.useState<string>("");
 
   const urlReferralAddress = router.getReferrerParameter();
-  const [storedReferralAddress, setStoredReferralAddress] = React.useState<string>("");
+  const [storedReferralAddress, setStoredReferralAddress] = React.useState<string | null>(null);
   const apiReferrerAddress = leaderboardMyInfo?.referrerAddress || "";
 
   const generateReferralLink = React.useCallback((): string => {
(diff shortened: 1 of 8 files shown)


packages/web/src/hooks/common/use-referral.tsx
```diff
@@ -41,30 +46,27 @@ export const useReferral = () => {
...
     } catch {
       return null;
     }
-  }, []);
+  }, [account?.address]);
 
-  const saveToSessionStorage = React.useCallback(
-    (referrerAddress: string) => {
-      if (!account?.address) return;
+  const saveToSessionStorage = (referrerAddress: string) => {
```
notJoon, line 59: There are effects or callbacks that include `saveToSessionStorage` in their dependency array, but it's not memoized.

I think that every time the component re-renders, a new function will be created, causing the `useEffect` to run unnecessarily.

packages/web/src/hooks/common/use-referral.tsx
```diff
@@ -89,35 +93,37 @@ export const useReferral = () => {
...
-      setStoredReferralAddress(storedInfo.referrerAddress);
-    }
-  }, [account?.address, getStoredReferrerInfo]);
-
   // Handling URL parameters and session storage priorities
   React.useEffect(() => {
     // Rank 1: URL parameters
-    if (urlReferralAddress && isValidAddress(urlReferralAddress)) {
+    const isUrlReferralAddressEmpty = urlReferralAddress === null || urlReferralAddress === undefined;
+    if (!isUrlReferralAddressEmpty && isValidAddress(urlReferralAddress)) {
```
notJoon, line 100: This condition check seems like it would happen automatically when parsing the bech32 format in the `isValidAddress` function, but I'm not sure.

However, by using a nullish check, this can handle it more concisely as follows.

For example, like this:
```ts
const hasUrlReferralAddress = urlReferralAddress != null;
if (hasUrlReferralAddress && isValidAddress(urlReferralAddress)) {
```
This can checks for `both` null and `undefined` at once.

notJoon (changes requested):
That looks good. I've left a comment for the `saveToSessionStorage` function, please check it.

packages/web/src/hooks/common/use-referral.tsx
tfrg: Removed saveToSessionStorage from the dependency array of saveRefererAddress to avoid unnecessary function regeneration.

Optimised the saveReferrerAddress function by memoising only the saveReferrerAddress function to useCallback so that the function is only regenerated when the account?.address changes.

The saveToSessionStorage function is only used inside saveReferrerAddress, so it didn't need to be memoised separately.

Previously, when we memoised saveToSessionStorage as a useCallback, we encountered closure issues that prevented the referrer input from updating its state properly, so removing the useCallback was more effective.

packages/web/src/hooks/common/use-referral.tsx
tfrg: Thanks review,

I've improved the code in the null check part.

notJoon (approved):
LGTM 👍

PR #664, opened 2025-04-09
[GSW-2199] leaderboard QA

tfrg:
## Description
This PR refactors the LeaderboardTableRow component to optimize how display values are calculated and rendered

## Details
- Consolidated three separate useMemo hooks (displaySwapVolume, displayPositionValue, displayStakingValue) into a single displayValues object
- Modified the formatUsdValue function to include the dollar sign ($) formatting
- Removed redundant calculations that were previously duplicated across multiple hooks
- Improved code readability by using the calculated values directly in the render method


jinoosss (approved):
👍

PR #666, opened 2025-04-10
[GSW-2229] Fix sorting behavior for empty values in pool list

tfrg:
## Description
This PR improves the sorting functionality in the PoolListContainer component by properly handling empty values.

## Details
- Modified the sort function to place items with "-" values at the end of the list when sorting in descending order
- Applied this behavior to TVL, Volume, Fees, and APR columns
- Added documentation comments to improve code readability and maintainability
- Implemented proper value transformation for sorting to handle edge cases

### DIFF
--- a/packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
+++ b/packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
@@ -167,13 +167,11 @@ const PoolListContainer: React.FC = () => {
    * Sort filtered pools based on current sort option
    */
   const sortedPools = useMemo(() => {
-    const temp = [...filteredPools];
     if (!sortOption) {
-      return temp.sort((a, b) => -sortValueTransform(a.tvl) + sortValueTransform(b.tvl));
+      return [...filteredPools].sort((a, b) => -sortValueTransform(a.tvl) + sortValueTransform(b.tvl));
     }
 
-    temp.sort(getSortFunction(sortOption?.key || TABLE_HEAD.TVL, sortOption?.direction || "desc"));
-    return temp;
+    return [...filteredPools].sort(getSortFunction(sortOption?.key || TABLE_HEAD.TVL, sortOption?.direction || "desc"));
   }, [filteredPools, sortOption, getSortFunction]);
 
   /**
--- a/packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
+++ b/packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
@@ -53,9 +53,12 @@ const PoolListContainer: React.FC = () => {
   }, [isClickOutside, keyword]);
 
   const sortValueTransform = (value: string) => {
-    if (!value) return -1;
+    if (!value || value === "-") return -Infinity;
 
-    return Number(value);
+    const numericValue = value.replace(/[$,]/g, "");
+    const number = Number(numericValue);
+
+    return isNaN(number) ? -Infinity : number;
   };
 
   const filteredPoolType = useCallback((poolType: POOL_TYPE, incentivized: boolean) => {
@@ -70,7 +73,7 @@ const PoolListContainer: React.FC = () => {
     return true;
   }, []);
 
-  const sortedPoolListInfos = useMemo(() => {
+  const filteredPools = useMemo(() => {
     const temp = poolListInfos.filter(info => {
       if (keyword !== "") {
         return (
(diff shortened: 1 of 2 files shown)


packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
```diff
@@ -34,143 +34,172 @@ const PoolListContainer: React.FC = () => {
...
+      const multiplier = direction === "asc" ? 1 : -1;
+
+      switch (key) {
+        case TABLE_HEAD.POOL_NAME:
+          return (
+            multiplier *
+            (direction === "asc"
+              ? b.tokenA.name.localeCompare(a.tokenA.name)
+              : a.tokenA.name.localeCompare(b.tokenA.name))
+          );
```
notJoon, line 125: The implementation to prevent unnecessary re-rendering is very good. 👍 

However, in this case, I'm wondering if there's a specific reason for checking the direction condition again? I think, it seems better to apply the `multiplier` consistently, and if there is a reason for this approach, adding comments to this section would be helpful for future maintenance.

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
```diff
@@ -34,143 +34,172 @@ const PoolListContainer: React.FC = () => {
...
-    switch (poolType) {
-      case POOL_TYPE.INCENTIVIZED:
-        return incentivized === true;
-      case POOL_TYPE.NONE_INCENTIVIZED:
-        return incentivized === false;
-      default:
-        break;
-    }
+    if (poolType === POOL_TYPE.INCENTIVIZED) return incentivized === true;
+    if (poolType === POOL_TYPE.NONE_INCENTIVIZED) return incentivized === false;
```
notJoon, line 145: Using `switch` here is a more TypeScript-like approach. How about changing it like this? Also, we can simplify the unnecessary checks in the return statement as follows:

```ts
const filteredPoolType = useCallback((poolType: POOL_TYPE, incentivized: boolean) => {
  switch (poolType) {
    case POOL_TYPE.INCENTIVIZED: return incentivized;
    case POOL_TYPE.NONE_INCENTIVIZED: return !incentivized;
    default: return true;
  }
}, []);
```

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
```diff
@@ -34,143 +34,172 @@ const PoolListContainer: React.FC = () => {
...
+    const number = Number(numericValue);
 
-    return Number(value);
+    return isNaN(number) ? -Infinity : number;
   };
 
+  /**
+   * Create sort function based on column and direction
+   */
+  const getSortFunction = useCallback((key: TABLE_HEAD, direction: "asc" | "desc") => {
```
notJoon, line 114: Using enums instead of string literal types like `"asc" | "desc"` allows you to better utilize IDE features (like auto completion). It also prevents errors due to typos and enables more strict type checking.

notJoon (changes requested):
I've left behind some really minor improvements. Other than that, it looks good 👍

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
```diff
@@ -34,143 +34,172 @@ const PoolListContainer: React.FC = () => {
...
+      info.tokenB.name.toLowerCase().includes(searchTerm) ||
+      info.tokenA.symbol.toLowerCase().includes(searchTerm) ||
+      info.tokenB.symbol.toLowerCase().includes(searchTerm)
+    );
+  }, []);
+
+  /**
+   * Transform string values to numbers for sorting
+   */
   const sortValueTransform = (value: string) => {
```
notJoon, line 102: consider specifying the return type.
```suggestion
  const sortValueTransform = (value: string): number => {
```

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
tfrg: @notJoon 
The old code was using multiplier but also checking the direction condition again internally, which was unnecessary redundancy. The revised code keeps the localeCompare call consistent and improves the sort direction to be controlled by multiplier only.

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
tfrg: Changing the existing code to a switch statement and removing the unnecessary comparison operation is a very good improvement. I've modified the code as you suggested

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
tfrg: After reviewing the code, I realised that we are using the sort direction as a string literal type (‘asc’ | ‘desc’), which should be improved to an enum.

packages/web/src/layouts/earn/containers/pool-list-container/PoolListContainer.tsx
tfrg: Reflected

notJoon (approved):
LGTM

PR #672, opened 2025-04-11
[GSW-2235] Remove upper limit for price impact calculation

tfrg:
## Description
Modified price impact calculation to remove upper limit while maintaining the lower limit of -100%. Implemented useCallback for performance optimization.

## Details
- Modified `formatPriceImpact` function to only apply a lower limit (-100%)
- Removed previous upper limit restriction to allow displaying any positive price impact value
- Maintained decimal formatting to 2 decimal places for consistency
- Implemented `useCallback` for performance optimization to prevent unnecessary re-renders

### DIFF
--- a/packages/web/src/hooks/swap/data/use-swap-handler.tsx
+++ b/packages/web/src/hooks/swap/data/use-swap-handler.tsx
@@ -360,6 +360,14 @@ export const useSwapHandler = () => {
     return "NONE";
   }, [priceImpact]);
 
+  const formatPriceImpact = useCallback((priceImpact: BigNumber | null | undefined) => {
+    if (!priceImpact) return 0;
+
+    if (priceImpact.isLessThan(-100)) return -100;
+
+    return Number(priceImpact.toFixed(2));
+  }, []);
+
   const swapButtonState: SwapButtonStateType = useMemo(() => {
     if (!connectedWallet) {
       return "WALLET_LOGIN";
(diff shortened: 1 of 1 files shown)


packages/web/src/hooks/swap/data/use-swap-handler.tsx
```diff
@@ -360,6 +360,14 @@ export const useSwapHandler = () => {
     return "NONE";
   }, [priceImpact]);
 
+  const formatPriceImpact = useCallback((priceImpact: BigNumber | null | undefined) => {
+    if (!priceImpact) return 0;
+
+    if (priceImpact.isLessThan(-100)) return -100;
+
+    return Number(priceImpact.toFixed(2));
+  }, []);
```
notJoon, line 369: I think this can reduce the conditions, but it's not necessary to apply this change.

```suggestion
  const formatPriceImpact = useCallback((priceImpact: BigNumber | null | undefined) => {
    if (!priceImpact) return 0;
    return Math.max(-100, Number(priceImpact.toFixed(2)));
  }, []);
```

notJoon (approved):
LGTM 👍

packages/web/src/hooks/swap/data/use-swap-handler.tsx
tfrg: Thanks for the great idea.

PR #676, opened 2025-04-11
[GSW-2236, GSW-2237] Update Token Feature

tfrg:
## Description
Token-related feature change PR

## Details
- Created logic showing the token path as a common utility.
- Updated default token information to match backend data.


packages/web/src/utils/token-utils.ts
```diff
@@ -70,8 +71,8 @@ export function formatTokenBalanceDisplay(balance: string, connectedWallet: bool
  * @param nativeCoinText - The text to display for native coins
  * @returns Formatted token path
  */
-export function formatTokenPath(path: string, isNative: boolean, nativeCoinText: string): string {
-  if (isNative) return nativeCoinText;
+export function formatTokenPath(path: string, isNative: boolean): string {
+  if (isNative) return STATIC_TEXT.NATIVE_COIN;
 
   // Remove 'gno.land/' prefix if present
   return path.replace(/^gno\.land\//, "");
```
notJoon, line 78: how about using `formatTokenModelPath` function in here? seems duplicated

notJoon (changes requested):

packages/web/src/utils/token-utils.ts
tfrg: @notJoon 
TokenModel and TokenInfo have different interface structures, each with its own dedicated utility function. We separated them because they sometimes require different interfaces as parameters, and it's safer to keep them separate because they have different data handling and requirements.

dongwon8247:
@tfrg Can you also change `Gnoland` to `gno.land` in the Receive/Send modal and Assets on the Portfolio page? No more Gnoland, so you can search all pages and change it accordingly. Thanks.

notJoon (approved):
LGTM 👍

PR #689, opened 2025-04-22
[GSW-2259] Enhance EarnAddLiquidity

tfrg:
## Description
This PR addresses performance issues and navigation problems in the position creation flow by optimizing the useEffect hook that handles URL query parameters.

## Details
- Excessive re-rendering during position creation causing lag
- Page not navigating back after transaction completion
- URL parameters getting stuck with outdated values

## Changes
- Refactored the useEffect hook to use memoized values for computed parameters
- Removed unnecessary dependencies from the dependency array
- Extracted repeated logic into separate useMemo hooks
- Improved code readability and maintainability


dongwon8247 (approved):

notJoon (approved):
🚀

PR #690, opened 2025-04-23
[GSW-2266] Filter position rewards to include only claimable amounts

tfrg:
## Description
This PR implements a filtering mechanism for position rewards, ensuring that only rewards with claimable amounts greater than zero are included in the position data structure.

## Changes
- Add `isClaimableReward` utility function to validate if a reward has a positive claimable amount
- Use BigNumber library for more accurate numerical comparisons
- Replace manual filtering logic with the reusable utility function
- Update position objects to contain only rewards that are actually claimable

### DIFF
--- a/packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-detailed-position-card/MyDetailedPositionCard.tsx
+++ b/packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-detailed-position-card/MyDetailedPositionCard.tsx
@@ -37,6 +37,7 @@ import { makeRouteUrl } from "@utils/page.utils";
 import { formatTokenExchangeRate } from "@utils/stake-position-utils";
 import { isEndTickBy, tickToPrice, tickToPriceStr } from "@utils/swap-utils";
 import { makeDisplayTokenAmount } from "@utils/token-utils";
+import { isClaimableReward } from "@utils/reward-utils";
 
 import { DailyEarningTooltipContent, PositionAPRInfo } from "../stat-tooltip-contents/DailyEarningTooltipContent";
 import { BalanceTooltipContent, PositionBalanceInfo } from "./BalanceTooltipContent";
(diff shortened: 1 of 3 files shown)


packages/web/src/utils/reward-utils.ts
```diff
@@ -37,3 +40,14 @@ export const isInternalRewardType = (rewardType: RewardType): boolean => {
...
+/**
+ * Functions to check reward claimability
+ * Returns true if the reward's claimable amount is greater than 0, otherwise false
+ * @param reward
+ * @returns boolean
+ */
+export const isClaimableReward = (reward: RewardModel): boolean => {
+  const claimableAmount = BigNumber(reward.claimableAmount || 0);
+  return !claimableAmount.isNaN() && claimableAmount.isGreaterThan(0);
+};
```
notJoon, line 53: When `BigNumber` has failed to convert a number, it returns `NaN`. Since `NaN` values will always return `false` when used with `isGreaterThan(0)`. So, there is no need to check for `NaN` separately.
```suggestion
export const isClaimableReward = (reward: RewardModel): boolean => {
  const claimableAmount = BigNumber(reward.claimableAmount ?? 0);
  return claimableAmount.isGreaterThan(0);
};
```

packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-detailed-position-card/MyDetailedPositionCard.tsx
```diff
@@ -338,16 +339,23 @@ const MyDetailedPositionCard: React.FC<MyDetailedPositionCardProps> = ({
...
+  }, [position]);
+
   const isClaimable = useMemo(() => {
-    if (position.rewards.length === 0) {
+    if (positionWithClaimableRewards.rewards.length === 0) {
       return false;
     }
 
-    return position.rewards.some(reward => {
+    return positionWithClaimableRewards.rewards.some(reward => {
```
notJoon, line 354: The reward already filtered using `isClaimableReward` in `positionWithClaimableRewards`.

I think it appears sufficient to simply check if there are any filtered rewards in `isClaimable`.

notJoon (changes requested):

packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-detailed-position-card/MyDetailedPositionCard.tsx
tfrg: Oh, right, I'll fix that.

PR #691, opened 2025-04-23
[GSW-2267] Portfolio: Implement ClaimAll functionality and enhance RewardTooltip

tfrg:
## Description
This PR implements the ClaimAll functionality in the Portfolio section and enhances the RewardTooltip component to improve user experience when claiming rewards.

## Details
- Added ClaimAll button in the Portfolio wallet balance detail section
- Refactored RewardTooltipContent component for better performance and maintainability
- Updated RewardModel to use RewardTokenModel for better type safety
- Unified rounding policies across reward calculations
- Enhanced display logic for claimable rewards
- Optimized reward data processing with useCallback and useMemo hooks

### DIFF
--- a/packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
+++ b/packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
@@ -23,112 +23,57 @@ export interface RewardTooltipContentProps {
   rewardInfo: { [key in DisplayRewardType]: PositionRewardForTooltip[] } | null;
 }
 
+interface RewardCategory {
+  type: DisplayRewardType;
+  rewards: PositionRewardForTooltip[] | null;
+  totalUSD: string;
+}
+
 const RewardTooltipContent: React.FC<RewardTooltipContentProps> = ({ rewardInfo }) => {
   const { getGnotPath } = useGnotToGnot();
   const { t } = useTranslation();
 
-  const swapFeeRewards = useMemo(() => {
-    if (!rewardInfo || rewardInfo.SWAP_FEE.length === 0) {
-      return null;
-    }
-    return rewardInfo.SWAP_FEE.sort((a, b) => (b.usd || 0) - (a.usd || 0));
-  }, [rewardInfo]);
+  const sortByUsd = React.useCallback(
+    (rewards: PositionRewardForTooltip[]) => rewards.sort((a, b) => (b.usd ?? 0) - (a.usd ?? 0)),
+    [],
+  );
 
-  const internalRewards = useMemo(() => {
-    if (!rewardInfo || rewardInfo.INTERNAL_REWARD.length === 0) {
-      return null;
-    }
-    return rewardInfo.INTERNAL_REWARD.sort((a, b) => (b.usd || 0) - (a.usd || 0));
-  }, [rewardInfo]);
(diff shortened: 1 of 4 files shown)


packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
```diff
@@ -22,140 +23,57 @@ export interface RewardTooltipContentProps {
...
-      return accum + current.usd;
-    }, null);
-    return formatOtherPrice(sumUSD, {
-      isKMB: false,
-    });
-  }, [externalRewards]);
+  const sortByUsd = React.useCallback(
+    (rewards: PositionRewardForTooltip[]) => [...rewards].sort((a, b) => (b.usd ?? 0) - (a.usd ?? 0)),
+    [],
+  );
```
notJoon, line 39: This function seems to be a simple sorting function with no dependencies.

If the component doesn't re-render frequently, using `useCallback` might be unnecessary optimization. Does it actually re-render frequently?

packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
```diff
@@ -22,140 +23,57 @@ export interface RewardTooltipContentProps {
...
 
-      return accum + current.usd;
-    }, null);
-    return formatOtherPrice(sumUSD, {
-      isKMB: false,
-    });
-  }, [noneRewards]);
+  const rewardsData = useMemo<RewardCategory[]>(() => {
+    const types: DisplayRewardType[] = ["SWAP_FEE", "INTERNAL_REWARD", "EXTERNAL_REWARD", "NONE"]; // Specify sort order
```
notJoon, line 68: Currently, the values of `DisplayRewardType` are hardcoded. Since `DisplayRewardType` is defined as a union type, if you're using TypeScript's type system, you could automatically extract all possible values.

Would you consider this approach if possible? Implementing this would ensure that when the type changes, the related constants are automatically updated, preventing mismatches between types and values in advance.

No need to apply this right away.

notJoon (approved):
The type definitions are clear and the logic is well-separated, looks very good.

I've left a few comments, but feel free to apply or just let it go.

packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
tfrg: We don't render often!

We'll strip off the useCallback later.

packages/web/src/components/common/reward-tooltip-content/RewardTooltipContent.tsx
```diff
@@ -22,140 +23,57 @@ export interface RewardTooltipContentProps {
...
-  const noneRewardUSD = useMemo(() => {
-    const isEmpty = !noneRewards || noneRewards?.length === 0;
+  const calculateTotalUsd = React.useCallback((rewards: PositionRewardForTooltip[] | null): string => {
+    if (!rewards || rewards.length === 0) return "-";
 
-    if (isEmpty) return "-";
+    const sumUSD = rewards.reduce(
+      (accum, current) => (current.usd !== null ? accum.plus(current.usd) : accum),
+      new BigNumber(0),
+    );
```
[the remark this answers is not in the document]
tfrg: Good, this makes the code more readable.

We will proceed this way in the future.

PR #693, opened 2025-04-28
hotfix: [GSW-2272] disabledBlackBars flag

tfrg:
## Description
This PR fixes the issue where colors are incorrectly applied outside the user-defined range in the liquidity pool chart. As shown in GSW-2272, colors should only appear within the range set by the user.

## Details
- Enhanced validation for maxTickPosition and minTickPosition values
- Added disableBlackBar flag to properly manage bar coloring
- Modified chart rendering logic to only display colors within the user-defined range


jinoosss (approved):
👍

PR #696, opened 2025-04-30
[GSW-2275] SendAmount

tfrg:
## Description
- This PR fixes a bug that causes the sendAmount to be set incorrectly when executing an IncreaseLiquidity transaction.

## Details
- Creating the `getSendAmount` Utility Function
- Common to sendAmount logic


jinoosss (approved):
👍

PR #703, opened 2025-06-04
[GSW-2274] Sorting Tooltip data

tfrg:
## Description
- This PR is an action that changes the alignment policy of tooltip data.

## Details
- IncentivizeToken Tooltip data
- Daily Earning (Token A/B)
- Reward (Token A/B)

### DIFF
--- a/packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-liquidity-content/MyLiquidityContent.tsx
+++ b/packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-liquidity-content/MyLiquidityContent.tsx
@@ -24,6 +24,7 @@ import { formatOtherPrice, formatPoolPairAmount } from "@utils/new-number-utils"
 import { makeDisplayTokenAmount } from "@utils/token-utils";
 import { mapToDisplayRewardType } from "@utils/reward-utils";
 import { DEFAULT_TOKEN_PRICE_RATIO } from "@common/values";
+import { sortTokensByPoolOrder } from "@utils/pool-utils";
 
 import { DailyEarningTooltipContent, PositionAPRInfo } from "../stat-tooltip-contents/DailyEarningTooltipContent";
 
@@ -184,6 +185,22 @@ const MyLiquidityContent: React.FC<MyLiquidityContentProps> = ({
         }
       });
 
+    if (positionData) {
+      Object.keys(infoMap).forEach(key => {
+        const rewardType = key as DisplayRewardType;
+        const rewardsArray = Object.values(infoMap[rewardType]);
+        if (rewardsArray.length > 0) {
+          const sortedRewards = sortTokensByPoolOrder(rewardsArray, positionData.tokenA.path, positionData.tokenB.path);
+
+          infoMap[rewardType] = {};
+
+          sortedRewards.forEach(reward => {
+            infoMap[rewardType][reward.token.priceID] = reward;
+          });
+        }
+      });
+    }
+
     return {
       SWAP_FEE: Object.values(infoMap["SWAP_FEE"]),
       INTERNAL_REWARD: Object.values(infoMap["INTERNAL_REWARD"]),
@@ -291,6 +308,20 @@ const MyLiquidityContent: React.FC<MyLiquidityContentProps> = ({
         const multipliedApr = categorizedMap[positionKey].apr;
         categorizedMap[positionKey].apr = multipliedApr ? Number(BigInt(multipliedApr) / totalLiquidity) / 1000 : null;
       });
+
+      if (positionData) {
+        const rewardType = typeKey as DisplayRewardType;
+        const rewardsArray = Object.values(categorizedMap);
+        if (rewardsArray.length > 0) {
+          const sortedRewards = sortTokensByPoolOrder(rewardsArray, positionData.tokenA.path, positionData.tokenB.path);
+
+          infoMap[rewardType] = {};
+
+          sortedRewards.forEach(reward => {
+            infoMap[rewardType][reward.token.priceID] = reward;
+          });
+        }
+      }
     });
 
     return {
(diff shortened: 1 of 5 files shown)


packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-liquidity-content/MyLiquidityContent.tsx
```diff
@@ -184,6 +185,22 @@ const MyLiquidityContent: React.FC<MyLiquidityContentProps> = ({
...
+      Object.keys(infoMap).forEach(key => {
+        const rewardType = key as DisplayRewardType;
+        const rewardsArray = Object.values(infoMap[rewardType]);
+        if (rewardsArray.length > 0) {
+          const sortedRewards = sortTokensByPoolOrder(rewardsArray, positionData.tokenA.path, positionData.tokenB.path);
+
+          infoMap[rewardType] = {};
+
+          sortedRewards.forEach(reward => {
+            infoMap[rewardType][reward.token.priceID] = reward;
```
notJoon, line 198: Could `reward.token.priceID` potentially be `undefined`?

packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-liquidity-content/MyLiquidityContent.tsx
```diff
@@ -291,6 +308,20 @@ const MyLiquidityContent: React.FC<MyLiquidityContentProps> = ({
...
+          });
+        }
+      }
```
notJoon, line 324: This part is exactly same as the commented above. how about extract it into a separate function?

packages/web/src/utils/pool-utils.ts
```diff
@@ -48,3 +49,27 @@ export function checkPoolStakingRewards(incentivized?: boolean) {
...
+    if (itemBCheckedPath === tokenACheckedPath) return 1;
+
+    // Then make tokens matching tokenB come next
+    if (itemACheckedPath === tokenBCheckedPath) return -1;
+    if (itemBCheckedPath === tokenBCheckedPath) return 1;
+
+    // Other tokens
+    return 0;
+  });
+}
```
notJoon, line 75: For sorting results to always be predictable, the comparator must satisfy _symmetry_ and _transitivity_. However, in the current function implementation, when both `a.token.path` and `b.token.path` are the same as `tokenAPath`, `compare(a, b)` (I'll temporarily name the function compare) returns `-1` because `itemACheckedPath === tokenACheckedPath`, and when calling `compare(b, a)`, it also returns `-1` because `itemBCheckedPath === tokenACheckedPath` in the same way.

In other words, when both items belong to priority A (or both belong to B), both `compare(a, b)` and `compare(b, a)` return `-1`, making it impossible to determine which one should come first. This can make the sorting results **non-deterministic** and lead to unexpected outcomes.

Additionally, the checkGnotPath function is being called multiple times. When `sort` sorts $N$ items, the number of internal comparisons is approximately $O(N log N)$. Since the same `tokenAPath` and `tokenBPath` are processed by `checkGnotPath` every time, this seems unnecessary, so it would be better to implement caching.

notJoon (changes requested):
mostly LGTM. left some modifications please revise these comments.

packages/web/src/layouts/pool/pool-detail/components/my-liquidity/my-liquidity-content/MyLiquidityContent.tsx
tfrg: Yes. We'll create a new `sortDisplayRewards` utility and manage it with common logic.

packages/web/src/utils/pool-utils.ts
tfrg: Yes. We will improve the sortTokensByPoolOrder function.

packages/web/src/utils/pool-utils.ts
```diff
@@ -48,3 +55,55 @@ export function checkPoolStakingRewards(incentivized?: boolean) {
...
+): T[] {
+  const tokenACheckedPath = checkGnotPath(tokenAPath);
+  const tokenBCheckedPath = checkGnotPath(tokenBPath);
+
+  return [...items].sort((a, b) => {
+    const itemACheckedPath = checkGnotPath(a.token.path);
+    const itemBCheckedPath = checkGnotPath(b.token.path);
+
+    const groupA = itemACheckedPath === tokenACheckedPath ? 0 : itemACheckedPath === tokenBCheckedPath ? 1 : 2;
+    const groupB = itemBCheckedPath === tokenACheckedPath ? 0 : itemBCheckedPath === tokenBCheckedPath ? 1 : 2;
```
notJoon, line 72: for more readability, it seems good to make as an function.

```suggestion
    const getGroup = (path: string): number => {
      if (path === tokenACheckedPath) return 0;
      if (path === tokenBCheckedPath) return 1;
      return 2;
    }

    const groupA = getGroup(itemACheckedPath);
    const groupB = getGroup(itemBCheckedPath);
```

packages/web/src/layouts/pool/pool-detail/components/staking/staking-content/incentivized-token-detail-tooltip-content/IncentivizeTokenDetailTooltipContent.tsx
```diff
@@ -19,65 +19,71 @@ const IncentivizeTokenDetailTooltipContent: React.FC<Props> = ({ poolStakings }:
...
-  }, []);
+  };
 
-  const incentiveTypeText = useCallback((type: INCENTIVE_TYPE) => {
+  const incentiveTypeText = (type: INCENTIVE_TYPE) => {
     return capitalize(type);
-  }, []);
+  };
+
+  const sortedStakings = React.useMemo(() => {
```
notJoon, line 30: It seems like it would be good to add guards since `NaN` could occur if `startTimestamp` is an invalid string, and `poolStaking` could potentially be `undefined`. 🤔

notJoon (approved):
The modified code looks practical. approve

packages/web/src/utils/pool-utils.ts
tfrg: Thanks, We'll refactor it later!

PR #712, opened 2025-07-03
[GSW-2323] Add onSuccess callback for confirmation-driven refetch in Tx events

tfrg:
## Description
This PR enhances the transaction event handling by introducing an explicit onSuccess callback that runs after the blockchain transaction is confirmed, decoupling it from the initial “emit” phase

## Details
- Executes only once the transaction status is SUCCESS, ensuring UI updates and data refetches occur after on-chain confirmation rather than immediately on emit.
- Reliable refetch with React Query invalidation

### DIFF
--- a/packages/web/src/utils/common.ts
+++ b/packages/web/src/utils/common.ts
@@ -31,7 +31,6 @@ export function wait<T>(runner: () => Promise<T>, waitTime = 1000, finishTime = 
   });
 }
 
-// common.ts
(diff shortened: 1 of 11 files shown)


packages/web/src/utils/common.ts
```diff
@@ -31,6 +31,11 @@ export function wait<T>(runner: () => Promise<T>, waitTime = 1000, finishTime =
   });
 }
 
+// common.ts
```
notJoon, line 34: leftover?
```suggestion
```

PR #724, opened 2025-07-25
[GSW-2338] Governance Page UI issue

tfrg:
## Description
This PR fixes UI issues on the Governance page.

## New Feature
- Proposal Drafts

## Details
- Governance Delegateable Quantity input unit needs to be modified
- Fix usd signs to be exposed equally
- (Mobile) Delegate Modal padding
- CreateProposalModal UI


packages/web/src/components/common/token-amount-input/TokenAmountInput.tsx
```diff
@@ -44,14 +46,35 @@ const TokenAmountInput: React.FC<TokenAmountInputProps> = ({
...
       if (value === "") {
         changeAmount("");
+        return;
       }
-      if (value !== "" && !isAmount(value)) return;
 
+      if (integersOnly) {
+        if (value.includes(".")) {
+          return;
+        }
```
notJoon, line 58: The decimal point is automatically filtered out by the regex `/^\d+$/`, making it an unnecessary check. It seems like it can be removed.

packages/web/src/hooks/governance/ui/use-proposal-draft.tsx
```diff
@@ -0,0 +1,115 @@
...
+  title: string;
+  description: string;
+}
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+interface UseProposalDraftProps {
+  setValue: UseFormSetValue<any>;
+  watch: UseFormWatch<any>;
+  isDirty: boolean;
+}
```
notJoon, line 22: It would be good to use concrete types instead of `any`. Looking at where `setValue` and `watch` are used, it seems like defining the following types would be better. We could probably use the `ProposalDraftData` type, but I'm not certain for this, so I created an arbitrary type. Please check and if `ProposalDraftData can be used`, please use that instead.

```typescript
interface ProposalFormValues {
  title: string;
  description: string;
}

interface UseProposalDraftProps {
  setValue: UseFormSetValue<ProposalFormValues>;
  watch: UseFormWatch<ProposalFormValues>;
  isDirty: boolean;
}
```

packages/web/src/layouts/governance/components/proposals-list/variable-select-box/VariableSelectBox.tsx
```diff
@@ -52,17 +58,36 @@ const VariableSelectBox: React.FC<VariableSelectBoxProps> = ({
...
+      if (
+        isOpen &&
+        selectRef.current &&
+        dropdownRef.current &&
+        !selectRef.current.contains(event.target as Node) &&
+        !dropdownRef.current.contains(event.target as Node)
+      ) {
+        onToggle(dropdownId, false);
+      }
+    };
```
notJoon, line 83: this condition is too complex. please consider breaking it down into separate functions

packages/web/src/layouts/governance/components/proposals-list/create-proposal-modal/CreateProposalModal.tsx
```diff
@@ -250,19 +279,37 @@ const CreateProposalModal: React.FC<CreateProposalModalProps> = ({
...
+  const shouldShowErrorText = (index: number, forMobile: boolean) => {
+    const fieldName = getFieldName(index, "param");
+    const hasError = paramErrors?.variable?.[index];
+    const isTouched = touchedFields.has(fieldName);
+
+    return (
+      isTouched &&
+      hasError &&
+      ((forMobile && breakpoint === DEVICE_TYPE.MOBILE) || (!forMobile && breakpoint !== DEVICE_TYPE.MOBILE))
+    );
```
notJoon, line 295: how about extracting the condition into a variable?

```typescript
const isCorrectDevice = forMobile === (breakpoint === DEVICE_TYPE.MOBILE);
return isTouched && hasError && isCorrectDevice;
```

packages/web/src/hooks/governance/ui/use-proposal-draft.tsx
tfrg: @notJoon 

Thank you for your review.
We've modified the code to reflect your feedback.

packages/web/src/hooks/governance/ui/use-proposal-draft.tsx
```diff
@@ -75,15 +73,14 @@ export function useProposalDraft({ setValue, watch, isDirty }: UseProposalDraftP
...
     const draft = loadProposalDraft();
     if (draft) {
-      setValue("title", draft.title);
-      setValue("description", draft.description);
+      setValue("title" as Path<T>, draft.title as T[Path<T>]);
+      setValue("description" as Path<T>, draft.description as T[Path<T>]);
     }
   }, [loadProposalDraft, setValue]);
+  const title = watch("title" as Path<T>) as string;
+  const description = watch("description" as Path<T>) as string;
```
notJoon, line 81: These type assertions could following compromise type safety:

1. Using `"title"` as `Path<T>` and `"description"` as `Path<T>` doesn't ensure these fields actually exist in type `T`. This could lead to runtime errors if the generic type doesn't contain these properties.

2. The assertion `draft.title as T[Path<T>]` completely bypass the language's type checking system, potentially allowing incompatible types to be assigned.

Consider constraining the generic type to ensure required fields exist:

```typescript
export function useProposalDraft<T extends FieldValues & ProposalDraftData>({
  setValue,
  watch,
  isDirty
}: UseProposalDraftProps<T>) {
  // ...
  useEffect(() => {
    const draft = loadProposalDraft();
    if (draft) {
      setValue("title", draft.title);
      setValue("description", draft.description);
    }
  }, [loadProposalDraft, setValue]);

  const title = watch("title");
  const description = watch("description");
  // ...
}
```
This will provides type safety and eliminates the need for unsafe type assertions.

tfrg: @notJoon 
We wanted to remove the type assertion completely, but due to the limitations of the Path<T> type system in react-hook-form, this was not possible.

Instead, we made the following improvements

Separated type assertions into constants to eliminate redundancy
Ensured compile-time safety with generic constraints
Explicitly stated why type assertions are needed with comments
This is a common pattern when using react-hook-form, and is a practical solution without compromising type safety.

1. TypeScript cannot infer that the string literal "title" is of type Path<T>.
2. We constrain T extends ProposalDraftData, but the compiler still cannot guarantee that "title" is a valid path to T is a valid path to T.
3. Path<T> in react-hook-form is a complex union type, cannot be directly matched to a string literal
4. Therefore, type assertions are unavoidable, and we've separated them into constants to ensure maximum type safety. to make it as type safe as possible.

```typescript
 const FORM_FIELDS = {
 title: "title" as Path<T>,
 description: "description" as Path<T>,
 } as const;
```

notJoon: seems reasonable lgtm

PR #726, opened 2025-07-30
[GSW-2337] Change Max to the maximum value of INT64

tfrg:
## Description
Changed the maximum tolerance value used when generating Approve transaction messages from `UINT64_MAX` to `INT64_MAX`

### DIFF
--- a/packages/web/src/common/clients/wallet-client/transaction-messages/common.ts
+++ b/packages/web/src/common/clients/wallet-client/transaction-messages/common.ts
@@ -4,7 +4,7 @@ import { GnoProvider } from "@common/clients/gno-provider/gno-provider";
 import { TransactionMessageError } from "@common/errors";
 import { DEFAULT_ALLOWANCE_LIMIT } from "@common/values";
 import { PACKAGE_NFT_PATH } from "@constants/environment.constant";
-import { MAX_INT64 } from "@utils/math.utils";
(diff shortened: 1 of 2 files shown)


packages/web/src/common/clients/wallet-client/transaction-messages/common.ts
```diff
@@ -150,8 +150,8 @@ export async function makeTransactionMessagesWithApproves(
 
     const previousAmount = accumulated[current.tokenPath][current.targetAddress].amount || "0";
     const sumAmountBN = BigNumber(previousAmount).plus(current.amount.toString());
-    if (sumAmountBN.isGreaterThan(MAX_UINT64.toString())) {
-      accumulated[current.tokenPath][current.targetAddress].amount = MAX_UINT64.toString();
+    if (sumAmountBN.isGreaterThan(MAX_INT64.toString())) {
+      accumulated[current.tokenPath][current.targetAddress].amount = MAX_INT64.toString();
```
notJoon, line 154: how about storing the result of `MAX_INT64.to_string()` in a variable to reduce unnecessary function calls and conversions? 

The practical approach would be to make it a global variable or constant.

notJoon (approved):
Left a minor improvements

notJoon (approved):
LGTM 👍

PR #731, opened 2025-08-12
feat: Enhance TypeScript Type Safety and Documentation in Utils

notJoon:
## Description

This PR significantly improves TypeScript type safety and adds comprehensive JSDoc documentation to the utility functions in `/packages/web/src/common/utils` package. 

#### 1. **Type Guards with Type Predicates**

```typescript
export const isNotNullOrUndefined = <T>(target: T | null | undefined): target is T => {
  return target !== null && target !== undefined;
};
```

**Why this pattern?**

TypeScript's structural type system doesn't automatically narrow types after runtime checks. Type predicates (`target is T`) explicitly tell the compiler that if the function returns `true`, the type should be narrowed within the conditional block. This bridges the gap between runtime behavior and compile-time type checking.

Example:

```typescript
// Without type predicate, TypeScript doesn't narrow the type
if (isNotNullOrUndefined(value)) {
  // value is still `T | null | undefined` here without type predicate
  // With type predicate, value is narrowed to `T`
  console.log(value.length); // Safe access to properties
}
```

#### 2. **Branded Types (Nominal Typing)**

```typescript
type TimezoneOffsetHours = number & { __brand: "TimezoneOffsetHours" };
type MinimalDenom = string & { __brand: "MinimalDenom" };
```

**Why this pattern?**

TypeScript uses structural typing by default, meaning types are compatible if their structures match. This can lead to errors where semantically different values are accidentally mixed. Branded types simulate nominal typing, making structurally identical types incompatible.

```typescript
// Without branding, these are interchangeable (bug-prone)
function convertCurrency(amount: number, rate: number) { /* ... */ }
convertCurrency(rate, amount); // Bug: arguments swapped, but no error

// With branding, compiler catches the mistake
type Amount = number & { __brand: "Amount" };
type Rate = number & { __brand: "Rate" };
function convertCurrency(amount: Amount, rate: Rate) { /* ... */ }
// convertCurrency(rate, amount); // error: cannot swap
```

#### 3. **Discriminated Unions (Tagged Unions)**

```typescript
type DateUtcToLocalResult =
  | { value: "-"; offsetHours: TimezoneOffsetHours; isEmpty: true }
  | { value: string; offsetHours: TimezoneOffsetHours; isEmpty: false };
```

**Why this pattern?**

Discriminated unions leverage TypeScript's ability to narrow types based on literal type fields. The discriminant field (`isEmpty`) acts as a runtime tag that TypeScript uses for exhaustive type checking and safe property access.

Example:
```typescript
function handleResult(result: DateUtcToLocalResult) {
  if (result.isEmpty) {
    // TypeScript knows result.value is "-" here
    console.log("Empty date");
  } else {
    // TypeScript knows result.value is string (not "-") here
    console.log(result.value.toUpperCase()); // Safe string operations
  }
}
```

### DIFF
--- a/packages/web/src/common/utils/data-check-util.ts
+++ b/packages/web/src/common/utils/data-check-util.ts
@@ -27,8 +27,7 @@ type TypeName = (typeof TYPE_NAMES)[keyof typeof TYPE_NAMES];
  */
 export const getType = (target: unknown): TypeName => {
   const typeString = Object.prototype.toString.call(target);
-  const match = typeString.match(/\[object (\w+)\]/);
-  return (match?.[1] || "Unknown") as TypeName;
+  return typeString.slice(8, -1) as TypeName;
 };
 
 /**
--- a/packages/web/src/common/utils/data-check-util.ts
+++ b/packages/web/src/common/utils/data-check-util.ts
@@ -1,29 +1,107 @@
-export const getType = (target: any) => {
-  return Object.prototype.toString.call(target).slice(8, -1);
+const TYPE_NAMES = {
+  String: "String",
+  Number: "Number",
+  Boolean: "Boolean",
+  Array: "Array",
+  Object: "Object",
+  Null: "Null",
+  Undefined: "Undefined",
+  Function: "Function",
+  Date: "Date",
+  RegExp: "RegExp",
+} as const;
+
+type TypeName = (typeof TYPE_NAMES)[keyof typeof TYPE_NAMES];
+
+// Type-safe getType function with union return type
+export const getType = (target: unknown): TypeName => {
+  const typeString = Object.prototype.toString.call(target);
+  const match = typeString.match(/\[object (\w+)\]/);
+  return (match?.[1] || "Unknown") as TypeName;
 };
 
-export const emptyStrCheckAfterTrim = (str: string) => {
+// Explicit return type
+export const emptyStrCheckAfterTrim = (str: string): boolean => {
   return Boolean(str.trim());
 };
 
-export const nullOrUndefinedCheck = (target: any) => {
+// Type guard with proper type predicate
+export const isNotNullOrUndefined = <T>(target: T | null | undefined): target is T => {
   return target !== null && target !== undefined;
 };
 
-export const notNullStringType = (v: any) => {
-  return getType(v) === "String" && nullOrUndefinedCheck(v);
+// Type guard for string type
+export const isNotNullString = (value: unknown): value is string => {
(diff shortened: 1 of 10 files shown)


packages/web/src/layouts/terms/TermsLayout.tsx
notJoon, line 1: formatted

packages/web/src/models/account/mapper/account-mapper.ts
notJoon, line 1: function name has changed

packages/web/src/common/utils/data-check-util.ts
```diff
@@ -1,29 +1,201 @@
-export const getType = (target: any) => {
-  return Object.prototype.toString.call(target).slice(8, -1);
+/**
+ * Type name constants for ECMAScript built-in types
+ */
+const TYPE_NAMES = {
```
notJoon, line 4: This would be ok to simply declaring it as a string literal type. However, I defined it as a `Record` type because there might be cases where runtime lookups are needed.

packages/web/src/common/utils/data-check-util.ts
```diff
@@ -1,29 +1,201 @@
...
+ * @param target - The value to check
+ * @returns true if the value is neither null nor undefined
+ * @example
+ * const value: string | null = getValue();
+ * if (isNotNullOrUndefined(value)) {
+ *   // value is now typed as string
+ *   console.log(value.length);
+ * }
+ */
+export const isNotNullOrUndefined = <T>(target: T | null | undefined): target is T => {
```
tfrg, line 57: It's great that you implemented type narrowing using Type Predicates.

tfrg (approved):

packages/web/src/common/utils/data-check-util.ts
```diff
@@ -1,29 +1,201 @@
...
+ * @param target - The value to check
+ * @returns The type name as a string (e.g., "String", "Number", "Array")
+ * @example
+ * getType("hello") // returns "String"
+ * getType(123) // returns "Number"
+ * getType([]) // returns "Array"
+ */
+export const getType = (target: unknown): TypeName => {
+  const typeString = Object.prototype.toString.call(target);
+  const match = typeString.match(/\[object (\w+)\]/);
```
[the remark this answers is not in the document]
notJoon: fixed 👍  https://github.com/gnoswap-labs/gnoswap-interface/pull/731/commits/fb4c574b246dbcd8557ba5f356c0ed49fa065a96

tfrg (approved):

PR #742, opened 2025-09-03
[GSW-2354] NativeToken

tfrg:
## Description
This PR improves the logic for determining if a token is native.

### DIFF
--- a/packages/web/src/utils/token-utils.ts
+++ b/packages/web/src/utils/token-utils.ts
@@ -212,6 +212,11 @@ export const formatTokenModelPath = (token: TokenModel): string => {
   return token.path.replace(/^gno\.land\//, "");
 };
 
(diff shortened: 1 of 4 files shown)


packages/web/src/utils/token-utils.ts
```diff
@@ -212,6 +212,11 @@ export const formatTokenModelPath = (token: TokenModel): string => {
   return token.path.replace(/^gno\.land\//, "");
 };
 
+export const isNativeTokenPath = (token: TokenModel): boolean => {
+  if (token.path === GNOT_TOKEN.path) return true;
+  return false;
```
notJoon, line 217: we can make this more simpler way.
```suggestion
  return token.path === GNOT_TOKEN.path;
```

notJoon (approved):
LGTM

PR #829, opened 2026-02-25
fix: sanitize all `dangerouslySetInnerHTML` usage with `DOMPurify`

notJoon:
## Changes

- Add `sanitizeHtml` utility based on DOMPurify with whitelist-only approach
- wrap all `dangerouslySetInnerHTML` usage points with `sanitizeHtml` to block script/event handler injection

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized HTML sanitization across the app to prevent XSS: dynamic descriptions, messages, tooltips and liquidity/pool UI now render sanitized content to improve safety.

* **Tests**
  * Added comprehensive tests validating sanitization against common and edge-case attack vectors while ensuring legitimate translated and formatted content continues to display correctly.


junghoon-vans (changes requested):
.

packages/web/package.json
```diff
@@ -31,13 +31,15 @@
     "@hookform/resolvers": "3.3.2",
     "@noble/hashes": "1.3.3",
     "@tanstack/react-query": "4.26.1",
+    "@types/dompurify": "^3.2.0",
```
junghoon-vans: `@types/dompurify` is a type-only package not needed at runtime and should be moved to `devDependencies`. Keeping it in `dependencies` unnecessarily inflates the production bundle size.

```json
// Move to devDependencies
"@types/dompurify": "^3.2.0"
```

packages/web/src/utils/sanitize-html.ts
```diff
@@ -0,0 +1,14 @@
+import DOMPurify from "dompurify";
```
junghoon-vans, line 1: **[Critical] SSR Compatibility Issue**

`DOMPurify` requires browser DOM APIs (`window`, `document`). In a Next.js SSR environment (Node.js), no DOM is available, so `DOMPurify.sanitize()` returns the input string as-is without sanitizing — meaning XSS protection is entirely bypassed on the server.

While the real XSS risk is client-side (all components in this PR use hooks), serving unsanitized HTML during a Next.js Pages Router SSR pass is still a concern.

**Recommended fix:**

Option 1 — Replace with `isomorphic-dompurify` (works safely in both SSR and CSR):
```typescript
import DOMPurify from "isomorphic-dompurify";
```

Option 2 — Add an SSR guard manually:
```typescript
import DOMPurify from "dompurify";
import { JSDOM } from "jsdom";

const purify =
  typeof window !== "undefined"
    ? DOMPurify
    : DOMPurify(new JSDOM("").window as unknown as Window);

export function sanitizeHtml(dirty: string): string {
  if (!dirty) return "";
  return purify.sanitize(dirty, { ALLOWED_TAGS, ALLOWED_ATTR, ALLOW_DATA_ATTR: false });
}
```

Option 1 is simpler and easier to maintain.

packages/web/src/utils/sanitize-html.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const ALLOWED_TAGS = ["div", "span", "p", "b", "strong", "em", "br", "i", "u"];
```
junghoon-vans, line 3: Please verify whether the snackbar `description` field in `FailContent.tsx` and `SuccessContent.tsx` may contain HTML links. Since `<a>` is not in `ALLOWED_TAGS`, any `<a href="...">` links in the description HTML will be silently stripped.

If links are not used there, no problem. But if they are, you should allowlist `href` with URL scheme restrictions:

```typescript
const ALLOWED_TAGS = ["div", "span", "p", "b", "strong", "em", "br", "i", "u", "a"];
const ALLOWED_ATTR = ["class", "href", "target", "rel"];

// Add to sanitize options to harden target="_blank"
FORCE_ATTR: { rel: "noopener noreferrer" }
```

packages/web/src/utils/sanitize-html.spec.ts
```diff
@@ -0,0 +1,171 @@
+import { sanitizeHtml } from "./sanitize-html";
+
+// ──────────────────────────────────────────────
+// 1. Preserves allowed tags
+// ──────────────────────────────────────────────
+describe("sanitizeHtml — preserves allowed tags", () => {
+  it("should preserve div and span tags", () => {
+    const input = "<div>Swapping <span>100</span> <span>GNOT</span></div>";
+    expect(sanitizeHtml(input)).toBe(input);
```
junghoon-vans: DOMPurify's HTML parsing may differ subtly between jest (jsdom) and a real browser. Exact string equality assertions like `toBe(input)` can be brittle — DOMPurify may normalize attribute ordering or self-closing tags depending on the environment.

Using `toContain` is more resilient:
```typescript
// Current
expect(sanitizeHtml(input)).toBe(input);

// Improved
const result = sanitizeHtml(input);
expect(result).toContain("<div>");
expect(result).toContain("<span>100</span>");
expect(result).not.toContain("<script");
```

If these tests pass in the current jest/jsdom setup they are fine for now, but they may break with environment changes.

junghoon-vans (changes requested):
.

packages/web/src/utils/sanitize-html.ts
notJoon: The snackbar `content.description` does not use `<a>` tags. Links are rendered as JSX `<a>` elements, so there is no need to include them in the allowed tags list at this time.

packages/web/src/utils/sanitize-html.ts
notJoon: ~~It seems `isomorphic-dompurify` is more actively maintained. Will replace with that package.~~

Bundle size is excessively large. Keeping `dompurify` with an SSR guard is the better approach.

notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/829/commits/d42c1b5551776f2924a788954d5769530acda1f4

packages/web/package.json
notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/829/commits/d42c1b5551776f2924a788954d5769530acda1f4

packages/web/src/utils/sanitize-html.ts
```diff
@@ -0,0 +1,14 @@
...
+const ALLOWED_ATTR = ["class"];
+
+export function sanitizeHtml(dirty: string): string {
+  if (!dirty) return "";
+
+  return DOMPurify.sanitize(dirty, {
+    ALLOWED_TAGS,
+    ALLOWED_ATTR,
+    ALLOW_DATA_ATTR: false,
+  });
```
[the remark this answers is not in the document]
notJoon: I think we need to maintaining `dompurify` package. because `isomorphic-dompurify`'s dependency size is too big.
https://github.com/gnoswap-labs/gnoswap-interface/pull/829#discussion_r2850581421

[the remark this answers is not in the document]
notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/829/commits/d42c1b5551776f2924a788954d5769530acda1f4

packages/web/src/utils/sanitize-html.ts
notJoon: Note: All components that use sanitizeHtml are client components that use hooks (useTranslation, useGnoscanUrl, etc.), and since they return an empty string during SSR and render with sanitized HTML upon client hydration, the current implementation is sufficient.

Additionally, there is the issue that using jsdom adds a significant number of dependencies.

packages/web/src/utils/sanitize-html.spec.ts
notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/829/commits/24e979dbea42e0a2dd2696dec6d83c9084032b2c

packages/web/src/utils/sanitize-html.ts
```diff
@@ -0,0 +1,18 @@
...
+  ALLOWED_ATTR,
+  ALLOW_DATA_ATTR: false,
+};
+
+export function sanitizeHtml(dirty?: string | null): string {
+  if (!dirty) return "";
+
+  if (typeof window === "undefined") return "";
+
+  return DOMPurify.sanitize(dirty, PURIFY_CONFIG);
```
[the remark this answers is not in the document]
notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/829/commits/0ed778a0b5af053f58cd33f47c57943a08ec95fc

packages/web/src/utils/sanitize-html.ts
```diff
@@ -0,0 +1,18 @@
...
+const PURIFY_CONFIG = {
+  ALLOWED_TAGS,
+  ALLOWED_ATTR,
+  ALLOW_DATA_ATTR: false,
+};
+
+export function sanitizeHtml(dirty?: string | null): string {
+  if (!dirty) return "";
+
+  if (typeof window === "undefined") return "";
```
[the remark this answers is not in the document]
notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/829/commits/0ed778a0b5af053f58cd33f47c57943a08ec95fc

junghoon-vans (approved):

PR #850, opened 2026-04-15
fix(pool): show actual owned position count

junghoon-vans:
## Summary
- update pool detail to show the real owned position count from the positions API instead of the paginated slice length
- update pool add additional info to fetch the full pool-specific position set when the total exceeds the default page size so staked and unstaked counts are not capped at 20
- keep the change frontend-only across the affected pool detail and pool add containers

## Validation
- yarn workspace @gnoswap-labs/web build

## Summary by CodeRabbit

* **Refactor**
  * Improved position data loading mechanism to dynamically calculate position limits based on total available positions, enhancing performance and display accuracy for pool liquidity positions.

### DIFF
--- a/packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
+++ b/packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
@@ -52,9 +52,30 @@ const AdditionalInfoContainer: React.FC = () => {
     return pools.some(pool => pool.poolPath === poolPath);
   }, [poolPath, pools]);
 
+  const {
+    totalPositionCount,
+    loading: isLoadingTotalPositionCount,
+  } = usePositionData({
+    isClosed: false,
+    poolPath: poolPath || "",
+    limit: 1,
+    queryOption: {
+      enabled: !!poolPath,
+    },
+  });
+
+  const positionLimit = useMemo(() => {
+    if (!totalPositionCount) {
+      return 20;
+    }
+
+    return totalPositionCount;
+  }, [totalPositionCount]);
+
   const { positions, loading: isLoadingPosition } = usePositionData({
     isClosed: false,
     poolPath: poolPath || "",
+    limit: positionLimit,
     queryOption: {
       enabled: !!poolPath,
     },
@@ -105,7 +126,13 @@ const AdditionalInfoContainer: React.FC = () => {
       handleClickGotoStaking={handleClickGotoStaking}
       pool={data}
       poolPath={poolPath}
-      isLoadingPool={isLoadingRPCPoolInfo || isFetchingFeetierOfLiquidityMap || isLoadingPoolInfo || isLoadingPosition}
+      isLoadingPool={
+        isLoadingRPCPoolInfo ||
+        isFetchingFeetierOfLiquidityMap ||
+        isLoadingPoolInfo ||
+        isLoadingPosition ||
+        isLoadingTotalPositionCount
+      }
       isLoadingGraph={isLoadingPoolInfo}
       isReversed={isReversed}
     />
--- a/packages/web/src/layouts/pool/pool-detail/components/my-liquidity/MyLiquidity.tsx
+++ b/packages/web/src/layouts/pool/pool-detail/components/my-liquidity/MyLiquidity.tsx
@@ -18,6 +18,7 @@ interface MyLiquidityProps {
   isOtherPosition: boolean;
   openedPosition: PoolPositionModel[];
   closedPosition: PoolPositionModel[];
+  totalPositionCount: number;
   breakpoint: DEVICE_TYPE;
   connected: boolean;
   isSwitchNetwork: boolean;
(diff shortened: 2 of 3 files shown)


packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
```diff
@@ -52,9 +52,30 @@ const AdditionalInfoContainer: React.FC = () => {
...
+    poolPath: poolPath || "",
+    limit: 1,
+    queryOption: {
+      enabled: !!poolPath,
+    },
+  });
+
+  const positionLimit = useMemo(() => {
+    if (!totalPositionCount) {
+      return 20;
```
notJoon: It would be better to make this a constant (or use an existing one if it’s already defined).

packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
```diff
@@ -52,9 +52,30 @@ const AdditionalInfoContainer: React.FC = () => {
...
+    totalPositionCount,
+    loading: isLoadingTotalPositionCount,
+  } = usePositionData({
+    isClosed: false,
+    poolPath: poolPath || "",
+    limit: 1,
+    queryOption: {
+      enabled: !!poolPath,
+    },
+  });
```
notJoon, line 67: `usePositionData` internally runs `useMakePoolPositions`, so the current flow (probe with `limit: 1` -> derive `positionLimit` from `totalPositionCount` -> call `usePositionData` again) triggers two sequential network requests and two separate react query cache entries.

Could we reduce this to a single call with a large enough `limit` and then verify the response's `totalCount`? this should be future works

packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
```diff
@@ -52,9 +52,30 @@ const AdditionalInfoContainer: React.FC = () => {
...
+
+  const positionLimit = useMemo(() => {
```
notJoon, line 69: since there's no upper bound on this limit, it's possible to load all data at once if a pool contains a large amount of data. it might b worth considering adding a upper cap

packages/web/src/layouts/pool/pool-detail/components/my-liquidity/MyLiquidity.tsx
```diff
@@ -84,7 +86,7 @@ const MyLiquidity: React.FC<MyLiquidityProps> = ({
           isSwitchNetwork={isSwitchNetwork}
           address={address}
           addressName={addressName}
-          positionLength={showedPositions.length}
+          positionLength={totalPositionCount}
```
notJoon, line 94: prev `showdPositions` toggles between "opened only" and "opened + closed" based on `isShowClosePosition`, so `positionLength` reflected what was actually rendered.

However, the new `totalPositionCount` is returned without an `isClosed` filter. So the header count may include closed positions and positions from other pools, which diverges from the previous semantics. (it's also not scoped by `poolPath unlike `accountPositions`)

could we either pass `isClosed`/`poolPath` filters into `usePositionData` here, or branch the displayed count based on `isShowClosePosition`? worth double checking

notJoon (changes requested):
left some considerations. please take a look and apply them if appropriate. thanks!

packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
junghoon-vans: API modifications are absolutely necessary. What if we fetch the entire data set at once?

packages/web/src/layouts/pool/pool-detail/components/my-liquidity/MyLiquidity.tsx
junghoon-vans: The previous implementation was incorrect. This change returns the correct position quant

packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
junghoon-vans: I think this is something that needs to be decided at the policy level. After all, I don't think we can keep showing it indefinitely.

packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
junghoon-vans: As you mentioned, once the limit is set, we can simply query using that value. We will respond accordingly once the policy is finalized.

jinoosss (approved):

notJoon (approved):

PR #852, opened 2026-04-15
fix(swap): remove swap fee handling in exact out

jinoosss:
### Summary
This PR fixes exact-out swap handling by removing extra fee padding logic that was being applied in the frontend.  
The entered exact-out amount is now passed through consistently for route estimation and transaction submission.

### What changed
- Removed `swapFee` default usage from `useSwap` and eliminated `exactOutPadding` calculations.
- Stopped multiplying exact-out amounts during:
  - debounced amount processing
  - `sendExactOutSwapRoute` request payload generation
  - swap amount display/state derivation
- Kept exact-in behavior unchanged while making exact-out amount handling consistent end-to-end.
- Reordered imports in `swap-router-repository-impl.ts` (non-functional cleanup).

### Why
The router already handles swap fee mechanics internally, so applying additional frontend-side padding for exact-out introduced incorrect amount inflation and mismatch risk between user intent and executed transaction parameters.

## Summary by CodeRabbit

* **Refactor**
  * Simplified swap calculations for exact output trades by removing padding adjustments.
  * Streamlined swap hook configuration by removing optional swap fee parameters.

* **Chores**
  * Reorganized internal utility imports in swap router module.

### DIFF
--- a/packages/web/src/hooks/swap/data/use-swap.tsx
+++ b/packages/web/src/hooks/swap/data/use-swap.tsx
@@ -33,7 +33,7 @@ interface UseSwapProps {
   swapFee?: number;
 }
 
-export const useSwap = ({ tokenA, tokenB, direction, slippage, swapFee = 15 }: UseSwapProps) => {
+export const useSwap = ({ tokenA, tokenB, direction, slippage }: UseSwapProps) => {
   const { transactionService, swapRouterRepository, rpcProvider } = useGnoswapContext();
   const { getNextReferralAddress, nextReferralAddress } = useReferral();
   const { data: gasTokenPrice } = useGetTokenPrices(GasToken.path);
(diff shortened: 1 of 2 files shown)


packages/web/src/hooks/swap/data/use-swap.tsx
```diff
@@ -33,7 +33,7 @@ interface UseSwapProps {
   swapFee?: number;
```
junghoon-vans, line 33: `swapFee` is still exposed on UseSwapProps, but the hook no longer consumes it after this change. That leaves a silent no-op input in the API and can mislead callers into thinking fee configuration still affects ExactOut behavior. I’d remove the prop here as well, or explicitly mark it as deprecated if you need a transition period.

junghoon-vans (approved):

PR #855, opened 2026-04-16
refactor(web): refactor `GasInfo` into a discriminated union

notJoon:
## Description

This PR refactors `GasInfo` from a product type into a discriminated union to remove impossible states and make gas-related flows safer at the type level.

For example, `GasInfo` allowed contradictory states such as:
- `hasError = false` with an error message
- gas numeric fields present alongside an error state

This change removes this impossible combinations from the type system and makes downstream usage explicit.

It also centralizes gas info construction and updates gas consumers to read gas values only from the success branch.

## Changes

- convert `GasInfo` to a discriminated union with:
    - `status: "success"`
    - `status: "error"`
- remove the legacy `hasError` field
- add `buildGasInfo()` to centralize gas info creation
- add `getGasUsed()` to safely read `gasUsed` from `GasInfo`
- update gas estimation and network fee hooks to use the shared builder
- narrow gas reads across swap, pool, wallet, and modal flows

## Summary by CodeRabbit

* **Refactor**
  * Normalized and centralized gas estimation and gas-used handling across transaction flows for more consistent fee behavior.
* **Bug Fixes**
  * Network fee now only shown for successful estimates and avoids presenting misleading fee values on failures.
* **Tests**
  * Added tests covering gas-info construction and gas-used extraction to ensure correct fee/result handling.

### DIFF
--- a/packages/web/src/hooks/gas/build-gas-info.ts
+++ b/packages/web/src/hooks/gas/build-gas-info.ts
@@ -1,0 +1,40 @@
+import { GAS_WANTED_BUFFER_MULTIPLIER } from "@common/values";
+import { calculateAdjustedGasFee } from "@utils/gas-utils";
+
+import { GasInfo } from "./types";
+
+export function buildGasInfo(
+  resultGasUsed: { gasUsed: number; errorMessage: string | null } | null,
+  gasPrice: number,
+): GasInfo {
+  if (!resultGasUsed) {
+    return {
+      status: "error",
+      simulateErrorMessage: "",
+    };
+  }
+
+  if (resultGasUsed.errorMessage !== null) {
+    return {
+      status: "error",
+      simulateErrorMessage: resultGasUsed.errorMessage,
+    };
+  }
+
+  const { gasFee, gasUsed, gasWanted } = calculateAdjustedGasFee(
+    resultGasUsed.gasUsed,
+    gasPrice,
+    GAS_WANTED_BUFFER_MULTIPLIER,
+  );
+
+  return {
+    status: "success",
+    gasFee,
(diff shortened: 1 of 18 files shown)


packages/web/src/hooks/gas/build-gas-info.ts
```diff
@@ -0,0 +1,40 @@
...
+    status: "success",
+    gasFee,
+    gasUsed,
+    gasWanted,
+    gasPrice,
+  };
+}
+
+export const getGasUsed = (gasInfo: GasInfo | null | undefined): number =>
+  gasInfo?.status === "success" ? gasInfo.gasUsed : 0;
```
junghoon-vans, line 40: Nice cleanup overall. One thing still worth calling out is that `error` / `null` cases now collapse to `0` here, so callers can no longer distinguish "gas estimation failed" from the fallback value at the call site. I do not think this is a blocker for this refactor, but if the intended product behavior is to stop submission when simulation fails, that policy still looks implicit rather than enforced. Since this helper is now the single constructor/accessor path, a small unit test matrix for `null` / `error` / `success` would also help lock in the intended semantics.

notJoon: Added unit tests for `buildGasInfo`, `getGasUsed` covering error/null and success cases. https://github.com/gnoswap-labs/gnoswap-interface/pull/855/commits/cf7983917c8a4a68343441a169883e1696ac3503

I did not change submission policy in this PR. If we need to simulation failure to explicitly block submission, I think that should be handled as a follow up policy/UI change rather than folded into this refactor.

junghoon-vans (approved):

PR #856, opened 2026-04-16
fix(swap): display incentivize history

jinoosss:
### Summary
This PR fixes incentivize history data accuracy for swap providers by aligning staking API parameters, mapping missing fields, and normalizing displayed token amounts by token decimals.  
It also improves WUGNOT display handling so users see the expected token presentation in history entries.

### Changes
- Updated staking history API request in `PoolRepositoryImpl`:
  - Changed query from `provider` to `address` (`/staking?address=...`) to match backend contract.
- Extended staking response/model mapping:
  - Added `penaltyAmount` to `PoolStakingResponse`.
  - Mapped `unvestedAmount` from `penaltyAmount` with a safe fallback in `PoolStakingMapper`.
- Improved incentivize history container data normalization:
  - Converted `remainingAmount`, `incentivizedAmount`, `unvestedAmount`, and `depositGnsAmount` into display units before rendering.
  - Applied WUGNOT -> GNOT display transformation for reward token metadata in history.
- Updated history UI rendering:
  - Replaced fixed decimal formatting with token-decimal-aware formatting.
  - Displayed `unvestedAmount` value instead of `-`.
- Included minor import-order cleanup (no functional impact).

### Why
Users were seeing incorrect or incomplete incentivize history values due to API parameter mismatch and inconsistent amount/decimal handling in the UI layer.  
This change ensures fetched data maps correctly and displays consistently across reward token types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token amount formatting in pool incentivization history to use each token's actual decimals (including deposit amounts).
  * Fixed API request parameter used when fetching pool staking data by address.
  * Corrected display of wrapped/alternate token variants so reward tokens show the expected token details.

* **New Features**
  * Display previously hidden unvested amount in incentivization history by populating and showing the unvested value.

### DIFF
--- a/packages/web/src/layouts/pool/pool-incentivize/containers/incentivize-pool-history-container/IncentivizePoolHistoryContainer.tsx
+++ b/packages/web/src/layouts/pool/pool-incentivize/containers/incentivize-pool-history-container/IncentivizePoolHistoryContainer.tsx
@@ -28,9 +28,9 @@ const IncentivizePoolHistoryContainer = () => {
       return {
         ...item,
         rewardToken: displayToken,
-        remainingAmount: String(makeDisplayTokenAmount(item.rewardToken, item.remainingAmount || "0") ?? "0"),
-        incentivizedAmount: String(makeDisplayTokenAmount(item.rewardToken, item.incentivizedAmount || "0") ?? "0"),
-        unvestedAmount: String(makeDisplayTokenAmount(item.rewardToken, item.unvestedAmount || "0") ?? "0"),
+        remainingAmount: String(makeDisplayTokenAmount(displayToken, item.remainingAmount || "0") ?? "0"),
+        incentivizedAmount: String(makeDisplayTokenAmount(displayToken, item.incentivizedAmount || "0") ?? "0"),
+        unvestedAmount: String(makeDisplayTokenAmount(displayToken, item.unvestedAmount || "0") ?? "0"),
         depositGnsAmount: String(
           makeDisplayTokenAmount(GNS_TOKEN, (item as ExtendedPoolStakingModel).depositGnsAmount || "0") ?? "0",
         ),
--- a/packages/web/src/layouts/pool/pool-incentivize/containers/incentivize-pool-history-container/IncentivizePoolHistoryContainer.tsx
+++ b/packages/web/src/layouts/pool/pool-incentivize/containers/incentivize-pool-history-container/IncentivizePoolHistoryContainer.tsx
@@ -1,11 +1,20 @@
+import BigNumber from "bignumber.js";
+
 import { useWallet } from "@hooks/wallet/data/use-wallet";
 import { ExtendedPoolStakingModel } from "@models/pool/pool-staking";
+import { TokenModel } from "@models/token/token-model";
 import { useGetPoolStakingListByAddress } from "@query/pools/use-get-pool-staking-list-by-address";
 
 import { GNS_TOKEN } from "@common/values/token-constant";
 import { useGnotToGnot } from "@hooks/token/data/use-gnot-wugnot";
-import { makeDisplayTokenAmount } from "@utils/token-utils";
 import IncentivizePoolHistory from "../../components/incentivize-pool-history/IncentivizePoolHistory";
+
+// String/BigNumber-based shift so raw integer amounts keep full precision even beyond
+const shiftToDisplay = (token: Pick<TokenModel, "decimals">, amount: string) => {
+  const bn = new BigNumber(amount);
+  if (bn.isNaN()) return "0";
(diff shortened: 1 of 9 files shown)


notJoon (dismissed):

packages/web/src/models/pool/mapper/pool-staking-mapper.ts
```diff
@@ -10,6 +10,8 @@ export class PoolStakingMapper {
   public static fromResponse(poolStaking: PoolStakingResponse): PoolStakingModel {
     return {
       ...poolStaking,
+      incentiveId: poolStaking?.incentiveId || "",
+      unvestedAmount: poolStaking.penaltyAmount || "0",
```
notJoon, line 14: it currently adding a penalty amount to the unvested value in here. could you please double check it?

notJoon (commented):
I've left one item to check. if everything looks good after checking, it should be fine to be merge

packages/web/src/models/pool/mapper/pool-staking-mapper.ts
jinoosss: This is a mapping of the API response.

notJoon (dismissed):

packages/web/src/models/pool/mapper/pool-staking-mapper.ts
```diff
@@ -10,6 +10,8 @@ export class PoolStakingMapper {
   public static fromResponse(poolStaking: PoolStakingResponse): PoolStakingModel {
     return {
       ...poolStaking,
+      incentiveId: poolStaking?.incentiveId || "",
```
junghoon-vans: If `incentiveId` is missing here, we silently convert it to an empty string, but the history UI still enables Claim based only on `endTimestamp` and later submits `incentiveID: ""` in `useRemoveExternalIncentive`. That means this row can generate an invalid `EndExternalIncentive` request instead of being treated as non-claimable. Can we guard this earlier, for example by keeping it nullable and disabling or hiding Claim when the ID is absent?

packages/web/src/models/pool/mapper/pool-staking-mapper.ts
junghoon-vans: The PR description says this should map `unvestedAmount` from `penaltyAmount` with a fallback, but this implementation does not actually fall back to the existing `unvestedAmount` field. If `penaltyAmount` is empty while `unvestedAmount` is still populated for some responses, we will show `0` here and discard valid data.

packages/web/src/layouts/pool/pool-incentivize/containers/incentivize-pool-history-container/IncentivizePoolHistoryContainer.tsx
```diff
@@ -17,10 +19,23 @@ const IncentivizePoolHistoryContainer = () => {
...
+    .map(item => {
+      const isWugnot = item.rewardToken.path === wugnotPath;
+      const displayToken = isWugnot
+        ? { ...item.rewardToken, ...getGnotPath(item.rewardToken), path: gnot?.path || item.rewardToken.path }
+        : item.rewardToken;
+
+      return {
+        ...item,
+        rewardToken: displayToken,
+        remainingAmount: String(makeDisplayTokenAmount(displayToken, item.remainingAmount || "0") ?? "0"),
```
junghoon-vans: This now normalizes every displayed history amount through `makeDisplayTokenAmount(...)`, and that helper currently does `BigNumber(Number(amount))`. Converting raw integer token amounts through `Number(...)` can lose precision once values exceed JS safe-number range, so large incentives or remaining balances may be displayed inaccurately. It would be safer to keep this path string/BigNumber-based end to end.

packages/web/src/models/pool/mapper/pool-staking-mapper.ts
jinoosss: commit: 94f7729bd7c6718939cb0b5b38a381c1e9b6c9a1

packages/web/src/models/pool/mapper/pool-staking-mapper.ts
jinoosss: commit: 94f7729bd7c6718939cb0b5b38a381c1e9b6c9a1

packages/web/src/layouts/pool/pool-incentivize/containers/incentivize-pool-history-container/IncentivizePoolHistoryContainer.tsx
jinoosss: aaf3dce22317177b33d7bd90eef240819855dc91

junghoon-vans (approved):

PR #859, opened 2026-04-17
refactor(swap): model confirm modal state as discriminated union

notJoon:
## Description

- Convert `SwapConfirmModalState` from a nullable object shape to a discriminated union with `idle` and `ready` states
- Narrow `ConfirmSwapModal` at the component boundary so ready-only data is handled as non-null inside the rendered content
- Add regression coverage for idle and ready modal rendering

I did not change the state transition to force `{ status: "idle" }` when modal inputs become unavailable. That behavior was potentially ambiguous, so the existing semantics were preserved.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for swap confirmation modal with state validation and token/price rendering assertions.

* **Refactor**
  * Improved swap confirmation modal state management with explicit idle and ready states, ensuring modal only renders when data is available. Enhanced internal component structure and simplified null checks for better reliability.


notJoon:
replaced by #917

PR #861, opened 2026-04-17
feat: enhance LineGraph and TokenChart with y-axis range

jinoosss:
## Summary
This PR improves token price chart rendering by introducing a deterministic y-axis range model and passing explicit y-axis bounds into the graph renderer.

### What changed
- Added explicit `yAxisMin` / `yAxisMax` support from container to chart graph:
  - `TokenChartContainer` -> `TokenChart` -> `TokenChartGraph` -> `LineGraph`
- Updated `LineGraph` to honor external y-axis bounds when provided:
  - Maps points directly against `[yAxisMin, yAxisMax]`
  - Preserves previous behavior when bounds are not provided
- Added `fillAreaBelowLine` option in `LineGraph`:
  - Fills gradient from the line down to the chart bottom to avoid fill escaping the curve baseline
- Reworked token chart y-axis generation logic in `TokenChartContainer`:
  - Uses fixed label count (`6`)
  - Derives axis window from:
    - `targetMax = highest * 1.2`
    - `targetMin = lowest * 0.8`
  - Restricts scale units to allowed form: `{1..9} * 10^n`
  - Ensures stable, readable labels for low/high-price tokens
- Applied minor chart label spacing adjustment in `TokenChartGraph.styles.ts` for visual alignment with the new range behavior
- Included non-functional cleanup in `LeaderboardListHeaderContainer` (import ordering / ref cast formatting)

## Why
Previous token chart scaling depended on internal auto-padding and could produce inconsistent y-axis behavior across periods/tokens.  
This change makes axis bounds explicit and deterministic, improving readability and making rendered line/area behavior predictable.

## Summary by CodeRabbit

* **New Features**
  * Support custom Y-axis bounds (min/max) for charts and optional fill of the area beneath line graphs.

* **Style**
  * Increased Y-axis label vertical padding for improved readability.

* **Refactor**
  * Overhauled Y-axis scaling and label logic with smart snapping to standard intervals, improved handling of low-value data, and better edge-case behavior for empty or zero datasets.

### DIFF
--- a/packages/web/src/components/common/line-graph/LineGraph.tsx
+++ b/packages/web/src/components/common/line-graph/LineGraph.tsx
@@ -106,6 +106,9 @@ export interface LineGraphProps {
   popupYValueFormatter?: (value: string) => string;
   hasNoLabel?: boolean;
   referenceTimeForFirstLine?: string;
+  yAxisMin?: string;
+  yAxisMax?: string;
+  fillAreaBelowLine?: boolean;
 }
 
 export interface LineGraphRef {
@@ -174,6 +177,9 @@ const LineGraph: React.FC<LineGraphProps> = ({
   popupYValueFormatter,
   hasNoLabel = false,
   referenceTimeForFirstLine,
+  yAxisMin,
+  yAxisMax,
+  fillAreaBelowLine = false,
 }: LineGraphProps) => {
   const COMPONENT_ID = (Math.random() * 100000).toString();
   const [activated, setActivated] = useState(false);
@@ -213,6 +219,12 @@ const LineGraph: React.FC<LineGraphProps> = ({
     let minTime: number;
     let maxTime: number;
 
+    const hasExternalYRange =
+      yAxisMin !== undefined &&
+      yAxisMax !== undefined &&
+      !new BigNumber(yAxisMin).isNaN() &&
+      !new BigNumber(yAxisMax).isNaN();
+
     const mappedDatas = (() => {
       const newDatas = datas.map(item => ({
         value: new BigNumber(item.value).toNumber(),
@@ -222,8 +234,8 @@ const LineGraph: React.FC<LineGraphProps> = ({
       const values = newDatas.map(data => data.value);
       const times = newDatas.map(data => data.time);
 
-      minValue = Math.min(...values);
-      maxValue = Math.max(...values);
+      minValue = hasExternalYRange ? new BigNumber(yAxisMin as string).toNumber() : Math.min(...values);
+      maxValue = hasExternalYRange ? new BigNumber(yAxisMax as string).toNumber() : Math.max(...values);
       minTime = Math.min(...times);
       maxTime = Math.max(...times);
 
@@ -388,6 +400,14 @@ const LineGraph: React.FC<LineGraphProps> = ({
     setBaseLineNumberWidth(baseLineNumberWidthComputation);
 
     const optimizeValue = function (value: number, height: number) {
+      if (hasExternalYRange) {
+        if (minMaxGap.isZero()) {
+          return height / 2;
+        }
+
+        return height - ((value - minValue) * height) / minMaxGap.toNumber();
+      }
+
       // The base line wrapper will > top and bottom of graph 10 % so the height will be 110% of graph height
       const graphHeight = (() => {
         if (showBaseLine) {
@@ -563,6 +583,21 @@ const LineGraph: React.FC<LineGraphProps> = ({
       return undefined;
     }
 
+    // When enabled, the gradient area fills from the price line down to the bottom of
+    // the chart so the fill never escapes the curve.
+    if (fillAreaBelowLine) {
+      const bottomY = height + (customHeight || 0);
+      let path = `M ${points[0].x},${bottomY}`;
+      path += ` L ${points[0].x},${points[0].y}`;
+      for (let i = 1; i < points.length; i++) {
+        const point = points[i];
+        path += smooth ? " " + bezierCommand(point, i, points) : ` L ${point.x},${point.y}`;
+      }
+      path += ` L ${points[points.length - 1].x},${bottomY}`;
+      path += " Z";
+      return path;
+    }
+
     // Start at the first point's x coordinate at firstPoint.y level
     let path = `M ${points[0].x},${firstPoint.y}`;
 
@@ -589,7 +624,7 @@ const LineGraph: React.FC<LineGraphProps> = ({
     path += " Z";
 
     return path;
-  }, [points, smooth, firstPoint.y]);
+  }, [points, smooth, firstPoint.y, fillAreaBelowLine, height, customHeight]);
 
   const isLightTheme = theme.themeKey === "light";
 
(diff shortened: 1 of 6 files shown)


packages/web/src/components/common/line-graph/LineGraph.tsx
```diff
@@ -388,6 +400,14 @@ const LineGraph: React.FC<LineGraphProps> = ({
     setBaseLineNumberWidth(baseLineNumberWidthComputation);
 
     const optimizeValue = function (value: number, height: number) {
+      if (hasExternalYRange) {
+        if (minMaxGap.isZero()) {
+          return height / 2;
+        }
+
+        return height - ((value - minValue) * height) / minMaxGap.toNumber();
+      }
```
junghoon-vans, line 412: The new `hasExternalYRange` branch still reuses `minMaxGap`, but that value is derived from the legacy internal-scaling path rather than from the explicit external bounds. In particular, this can misplace points when the dataset has a single point or when a future caller passes a negative `yAxisMin`, because the denominator can collapse to `yAxisMax` instead of `yAxisMax - yAxisMin`. Since this PR expands `LineGraph` into a more general API, I think this branch should compute its own gap directly from the external range to avoid latent coordinate bugs.

packages/web/src/layouts/token-detail/containers/token-chart-container/TokenChartContainer.tsx
```diff
@@ -263,71 +316,104 @@ const TokenChartContainer: React.FC = () => {
...
+  ): { labels: string[]; minValue: string; maxValue: string } => {
+    const Y_AXIS_LABEL_COUNT = 6;
+    const Y_AXIS_STEP_COUNT = Y_AXIS_LABEL_COUNT - 1;
+    const useSignificantFormat = (tokenPrice ?? 0) < 1;
+
+    const formatYAxisValue = (value: BigNumber) => {
+      if (useSignificantFormat) {
+        return truncateToSignificantDigits(value, 3);
+      }
+      return value.toFixed(2);
```
junghoon-vans: This formatting rule can regress label readability for tokens priced above 1 USD but moving within a very narrow range. In that case the chart step can be much smaller than `0.01`, while `toFixed(2)` rounds multiple distinct axis values into the same displayed label. The previous implementation adapted precision to the actual range, so I think the new deterministic scaling should still preserve range-sensitive formatting instead of locking all `>= 1` tokens to two decimals.

packages/web/src/layouts/token-detail/containers/token-chart-container/TokenChartContainer.tsx
```diff
@@ -263,71 +316,104 @@ const TokenChartContainer: React.FC = () => {
...
-      });
+  }, [currentTab, chartData, countXAxis, breakpoint, currentPrice]);
+
+  const getYAxisInfo = (
+    datas: string[],
+    tokenPrice?: number,
+  ): { labels: string[]; minValue: string; maxValue: string } => {
+    const Y_AXIS_LABEL_COUNT = 6;
+    const Y_AXIS_STEP_COUNT = Y_AXIS_LABEL_COUNT - 1;
+    const useSignificantFormat = (tokenPrice ?? 0) < 1;
```
junghoon-vans: There is also a small loading-state UX issue here: while `tokenPrice` is still `undefined`, this expression evaluates to `true`, so high-priced tokens briefly render with significant-digit formatting and then switch to fixed decimals once the price query resolves. It is not a functional bug, but the Y-axis format can visibly flicker during loading.

packages/web/src/layouts/token-detail/containers/token-chart-container/TokenChartContainer.tsx
```diff
@@ -263,71 +316,104 @@ const TokenChartContainer: React.FC = () => {
...
-    const maxPoint = maxValue.plus(originalGap.multipliedBy(0.05));
-    const dynamicPrecision = Math.min(6, Math.max(4, (maxPoint.minus(minPoint).decimalPlaces() ?? 0) + 2));
-    const formatYAxisValue = (value: BigNumber, precision = dynamicPrecision) =>
-      formatPrice(value, {
-        usd: false,
-        isKMB: false,
-        lessThan1Significant: precision,
-        greaterThan1Decimals: precision,
-      });
+  }, [currentTab, chartData, countXAxis, breakpoint, currentPrice]);
```
junghoon-vans: `getChartInfo` is wrapped in `useCallback`, but it is invoked immediately during render via `chartInfo={getChartInfo()}`. That means the full computation still runs on every render, so the callback identity is stable without actually memoizing the result. This looks like a better fit for `useMemo`, especially now that the Y-axis preparation path has become heavier.

packages/web/src/components/common/line-graph/LineGraph.tsx
jinoosss: 9fd1135ebd664264224a83d71be05411477b5db5

packages/web/src/layouts/token-detail/containers/token-chart-container/TokenChartContainer.tsx
jinoosss: 9fd1135ebd664264224a83d71be05411477b5db5
```
  const formatYAxisValue = (value: BigNumber, scaleUnit?: BigNumber) => {
    if (useSignificantFormat) {
      return truncateToSignificantDigits(value, 3);
    }
    const unitDecimals = scaleUnit ? Math.max(0, -(scaleUnit.e ?? 0)) : 0;
    const decimals = Math.max(2, unitDecimals);
    return value.toFixed(decimals);
  };
```

packages/web/src/layouts/token-detail/containers/token-chart-container/TokenChartContainer.tsx
jinoosss: 9fd1135ebd664264224a83d71be05411477b5db5
```
const useSignificantFormat = tokenPrice !== undefined && tokenPrice < 1;
```

packages/web/src/layouts/token-detail/containers/token-chart-container/TokenChartContainer.tsx
jinoosss: 9fd1135ebd664264224a83d71be05411477b5db5
```
[currentTab, chartData, countXAxis, breakpoint, currentPrice, size.width]
```

junghoon-vans (dismissed):

packages/web/src/components/common/line-graph/LineGraph.tsx
```diff
@@ -174,6 +177,9 @@ const LineGraph: React.FC<LineGraphProps> = ({
   popupYValueFormatter,
   hasNoLabel = false,
   referenceTimeForFirstLine,
+  yAxisMin,
+  yAxisMax,
+  fillAreaBelowLine = false,
```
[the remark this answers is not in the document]
junghoon-vans: I agree this should be fixed before merge. `updatePoints` reads `yAxisMin`/`yAxisMax`, so the effect deps should include both values to avoid stale point/baseline rendering when only the external range changes.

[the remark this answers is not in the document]
jinoosss: 72a9fd00be9a915b9e21594441de5a6fe29177b4

junghoon-vans (approved):

PR #862, opened 2026-04-18
fix(pool): use distributable amount in staking views

junghoon-vans:
## Summary
- switched the pool staking frontend model and converter to consume `distributableAmount`
- updated the staking tooltip and incentivized pool history views to display `distributableAmount` instead of `remainingAmount`
- kept compatibility with the API response while moving active UI usage away from deprecated staking fields

## Testing
- corepack yarn workspace @gnoswap-labs/web build

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated pool staking reward amounts displayed in tooltips and transaction history views to accurately reflect distributable quantities across incentivized pools.


junghoon-vans:
Closing this PR because the rollout was simplified to an API-only change. The frontend can keep using remainingAmount, so this branch is no longer needed for the current direction.

PR #876, opened 2026-05-13
fix(wallet): handle locked Adena login

junghoon-vans:
## Summary
- Treat Adena `WALLET_LOCKED` responses as a user-action-required state instead of a generic connection failure.
- Prevent the Gnoswap connection failure modal from opening on top of Adena's locked-wallet prompt.

## Changes
- Added explicit locked-wallet checks for both `addEstablishedSite()` responses and `getAccount()` errors.
- Reset `loadingConnect` to `initial` for locked Adena flows so the login UI returns to idle without showing an extra failure modal.
- Preserve the existing failure modal behavior for non-lock wallet connection errors.

## Test Coverage
- LSP diagnostics on `packages/web/src/hooks/wallet/data/use-wallet.ts`: passed
- `yarn workspace @gnoswap-labs/web lint`: passed with existing warnings
- `yarn workspace @gnoswap-labs/web build`: passed

## Summary by CodeRabbit

* **New Features**
  * Prevents concurrent wallet connection attempts for more reliable connects.
  * Allows targeting a specific wallet client when switching networks or connecting.

* **Bug Fixes**
  * Improved detection and handling of locked wallets to avoid unnecessary errors and reset connection UI state.
  * Ensures account data is retrieved and persisted after successful connection.

* **Tests**
  * Added unit tests covering locked-wallet detection logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnoswap-labs/gnoswap-interface/pull/876)

### DIFF
--- a/packages/web/src/hooks/wallet/data/use-wallet.ts
+++ b/packages/web/src/hooks/wallet/data/use-wallet.ts
@@ -1,11 +1,11 @@
 import { WalletClient } from "@common/clients/wallet-client";
 import { AdenaClient } from "@common/clients/wallet-client/adena";
-import { AdenaError, ERROR_VALUE } from "@common/errors/adena";
 import { NetworkData } from "@constants/chains.constant";
 import { DEFAULT_CHAIN_ID, SUPPORT_CHAIN_IDS } from "@constants/environment.constant";
 import { AUTH_STORE_KEY } from "@hooks/common/use-auto-disconnect";
 import { useGnoswapContext } from "@hooks/common/use-gnoswap-context";
 import { useSocialWalletContext } from "@hooks/common/use-social-wallet-context";
+import { AccountMapper } from "@models/account/mapper/account-mapper";
 import { useGetTokenBalancesFromChain } from "@query/address";
 import {
   ACCOUNT_SESSION_INFO_KEY,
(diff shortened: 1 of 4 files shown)


packages/web/src/hooks/wallet/data/use-wallet.ts
```diff
@@ -191,35 +199,50 @@ export const useWallet = () => {
...
+        setLoadingConnect("initial");
         return;
       }
 
-      const established = await accountRepository.addEstablishedSite().catch(() => null);
+      const established = await currentWalletClient.addEstablishedSite("Gnoswap");
 
-      if (established === null) {
+      if (isWalletLockedResponse(established)) {
+        setLoadingConnect("initial");
```
jinoosss, line 227: This may be unnecessary.

jinoosss (approved):

packages/web/src/hooks/wallet/data/use-wallet.ts
junghoon-vans: I think this is still needed. `addEstablishedSite()` returns the raw `WalletResponse` and does not run through `AdenaError.validate()`, so `WALLET_LOCKED` can arrive on this response path rather than the catch path. Without this branch, `code=2000/type=WALLET_LOCKED` would fall through to `setLoadingConnect("error")` and open the generic failure modal instead of returning the connect UI to idle.

PR #880, opened 2026-05-14
fix: refetch bins on liquidity changes and debounce pool state

jinoosss:
### Summary

Improves how pool liquidity bins are fetched and keyed so the pool detail graph stays consistent when liquidity or the current tick changes. Adds debounced pool-state inputs to `useGetBinsByPath`, extends the query key with deposited token amounts, and wires those values from `PoolPairInformationContainer` so bins refetch when on-chain pool balances update—not only when the tick changes.

### Problem

- Bins are centered on a **server tick** at fetch time; mixing **bins** from one response with **currentTick** from another source or moment skews the chart.
- After liquidity changes, pool detail can refetch with new balances while bins were still keyed mainly on path/tick, so the UI could lag or double-fetch awkwardly on first paint.

### Solution

1. **`useGetBinsByPath`**
   - Accepts optional `poolDepositedTokenAAmount` and `poolDepositedTokenBAmount`.
   - Debounces `{ currentTick, token A/B balances }` (default **1000 ms**) so refetches wait for the indexer to settle after state changes.
   - Builds `queryKey` from caller prefix + **debounced** tick and balances so liquidity changes invalidate and refetch bins.
   - Enables the query only when **`debouncedCurrentTick` is defined**, avoiding a fetch keyed on stale/undefined tick before debounce settles.
   - Returns **`pairedTick`** aligned with the tick used for that bins payload (works with `keepPreviousData` so tick and bins update together).

2. **`PoolPairInformationContainer`**
   - Passes `data?.tokenABalance` / `data?.tokenBBalance` into the hook.
   - Uses a shorter `queryKey` prefix; the hook appends debounced dependencies.
   - `enabled` is `!!poolPath` only; gating uses debounced tick inside the hook.
   - Adds a **60s** `refetchInterval` as a slow fallback if balances lag.
   - Pool object passed downstream uses **`pairedTick ?? pool currentTick`** so the displayed tick always matches the bins being shown.

3. **Call sites**
   - `useSelectPool` and `IncentivizedPoolCard` pass explicit `null` for the new balance args where bins do not depend on pool-detail liquidity.

## Summary by CodeRabbit

* **Refactor**
  * Debounced pool/position inputs to reduce unnecessary data refetches and stabilize graphs.
  * Query keys and enablement logic refined to align refetches with relevant pool-detail changes.

* **Breaking Changes**
  * Several hooks now accept optional pool/position state inputs (debounced) which affect when bins data is requested; callers should pass context where applicable.

* **Style**
  * Minor import and call-site argument reorderings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnoswap-labs/gnoswap-interface/pull/880)

### DIFF
--- a/packages/web/src/components/common/my-position-card-list/my-position-card/MyPositionCard.tsx
+++ b/packages/web/src/components/common/my-position-card-list/my-position-card/MyPositionCard.tsx
@@ -55,9 +55,19 @@ const MyPositionCard: React.FC<MyPositionCardProps> = ({
   const [isMouseoverGraph, setIsMouseoverGraph] = useState(false);
   const [shortenInRange, setShortenInRange] = useState(false);
 
-  const { data: bins40, isFetched: isFetchedBins } = useGetPositionBins(position.lpTokenId, 40, {
-    enabled: isMouseoverGraph,
-  });
+  const { data: bins40, isFetched: isFetchedBins } = useGetPositionBins(
+    position.lpTokenId,
+    40,
+    {
+      liquidity: position.liquidity,
+      poolCurrentTick: position.pool?.currentTick,
+      poolTokenABalance: position.pool?.tokenABalance,
+      poolTokenBBalance: position.pool?.tokenBBalance,
+    },
+    {
+      enabled: isMouseoverGraph,
+    },
+  );
 
   const { prefetch } = usePrefetchNavigation({
     pageType: "POOL",
--- a/packages/web/src/hooks/common/use-debounce.tsx
+++ b/packages/web/src/hooks/common/use-debounce.tsx
@@ -3,15 +3,31 @@ import React from "react";
 const useDebounce = <T,>(value: T, delay: number): T => {
   const [debouncedValue, setDebouncedValue] = React.useState<T>(value);
 
+  // Keep the latest `value` in a ref so the debounce effect doesn't have to
+  // depend on `value` by reference. A caller passing a fresh object literal
+  // every render would otherwise re-arm the timer endlessly (timer fires ->
+  // setState -> re-render -> new object -> timer re-arms = a permanent loop).
+  const valueRef = React.useRef(value);
+  valueRef.current = value;
+
+  // Drive the effect off a value-based key instead. For primitives this is the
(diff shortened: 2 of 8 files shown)


junghoon-vans (commented):
I found two major issues worth addressing before merge:

1. `useDebounce` is now called with a fresh object literal on every render. The existing debounce hook depends on `[value, delay]` by reference, so this can create a permanent 1s render loop: timer fires → `setDebouncedValue(newObject)` → render → new object → timer again. Could we debounce each primitive separately, or memoize the debounce input object before passing it to `useDebounce`?

2. Should position bins be refreshed through the same pool-state change path as pool bins? This PR updates `useGetBinsByPath`, but `useGetPositionBins` still keys only by `[QUERY_KEY.positionBins, lpTokenId, count]`. If liquidity changes while a position remains the same, the position bin graph may still show stale bin amounts until an explicit invalidation/remount happens. Could you check whether position bins need the same balance/tick-driven refresh or an explicit invalidation after liquidity mutations?

jinoosss:
> I found two major issues worth addressing before merge:
> 
> 1. `useDebounce` is now called with a fresh object literal on every render. The existing debounce hook depends on `[value, delay]` by reference, so this can create a permanent 1s render loop: timer fires → `setDebouncedValue(newObject)` → render → new object → timer again. Could we debounce each primitive separately, or memoize the debounce input object before passing it to `useDebounce`?
> 2. Should position bins be refreshed through the same pool-state change path as pool bins? This PR updates `useGetBinsByPath`, but `useGetPositionBins` still keys only by `[QUERY_KEY.positionBins, lpTokenId, count]`. If liquidity changes while a position remains the same, the position bin graph may still show stale bin amounts until an explicit invalidation/remount happens. Could you check whether position bins need the same balance/tick-driven refresh or an explicit invalidation after liquidity mutations?

I modified the system so that dependencies can be managed in a raw format, and I updated the position bins to track changes in the pool price and liquidity.

junghoon-vans (approved):
Thanks for addressing the review feedback. The latest commit resolves the debounce object identity issue and extends position bins refetching to depend on pool state changes.

PR #884, opened 2026-05-14
fix: reduce delay for data refresh across multiple components

jinoosss:


junghoon-vans (approved):
Reviewed the refresh timing changes. No blocking issues found; pending-event polling remains scoped to active transaction events, and CI is green.

PR #885, opened 2026-05-14
fix: enhance staking components and integrate video guide functionality

jinoosss:
## Summary by CodeRabbit

* **New Features**
  * Added video guide modal for staking guidance
  * Keyboard navigation support for pool selection (Enter/Space keys)

* **Improvements**
  * Enhanced staking pools table with hover effects and improved interactivity
  * Added navigation prefetch on pool row hover for faster page transitions
  * Updated default pool sorting behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnoswap-labs/gnoswap-interface/pull/885)

### DIFF
--- a/packages/web/src/layouts/earn/components/earn-my-positions/earn-my-positions-header/EarnMyPositionsHeader.tsx
+++ b/packages/web/src/layouts/earn/components/earn-my-positions/earn-my-positions-header/EarnMyPositionsHeader.tsx
@@ -3,13 +3,13 @@ import { useTranslation } from "react-i18next";
 
 import Button, { ButtonHierarchy } from "@components/common/button/Button";
 import Switch from "@components/common/switch/Switch";
+import { VIDEO_GUIDE_TYPES } from "@constants/video-guide.constant";
 import { useGnoscanUrl } from "@hooks/common/use-gnoscan-url";
 import { PoolPositionModel } from "@models/position/pool-position-model";
-import { VIDEO_GUIDE_TYPES } from "@constants/video-guide.constant";
 
-import { HeaderTextWrapper, PositionsWrapper } from "./EarnMyPositionsHeader.styles";
 import VideoGuideTrigger from "@components/common/video-guide-trigger/VideoGuideTrigger";
 import { useWindowSize } from "@hooks/common/use-window-size";
+import { HeaderTextWrapper, PositionsWrapper } from "./EarnMyPositionsHeader.styles";
 
 export interface EarnMyPositionsHeaderProps {
   address?: string | null;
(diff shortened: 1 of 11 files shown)


packages/web/src/layouts/pool/pool-stake/containers/available-staking-pools-container/AvailableStakingPoolsContainer.tsx
```diff
@@ -64,7 +63,7 @@ const AvailableStakingPoolsContainer: React.FC = () => {
 
   const onSelectPool = useCallback(
     (selectedPoolPath: string) => {
-      router.movePageWithPoolPath("POOL_STAKE", selectedPoolPath);
+      router.movePageWithPoolPath("POOL", selectedPoolPath, "staking");
```
junghoon-vans, line 71: Changing this from `/earn/pool/stake?poolPath=...` to `/earn/pool?poolPath=...#staking` can fail to scroll users to the staking section when they have no positions in that pool. In `PoolDetail`, the scroll effect returns early with `window.scrollTo({ top: 0 })` when `positions.length === 0`, so it never reaches the `hash === "staking"` branch. If we keep the `POOL#staking` navigation, we should also move the staking-hash handling before that empty-positions guard.

packages/web/src/layouts/pool/pool-stake/containers/available-staking-pools-container/AvailableStakingPoolsContainer.tsx
```diff
@@ -49,12 +49,11 @@ const AvailableStakingPoolsContainer: React.FC = () => {
 
   const onChangeSort = useCallback((key: AvailableStakingPoolsSortKey) => {
     setSort(prev => {
-      // Different column: start at ASC.
+      // Different column: start at DESC.
       if (prev.key !== key) {
-        return { key, direction: "asc" };
+        return { key, direction: "desc" };
```
junghoon-vans, line 59: Now that the Pool Name column starts sorting in descending order on the first click, this mismatch is more visible: the `poolName` comparator sorts by `poolPath`, but the rendered column shows the token pair and fee. Users will expect the sort order to follow the displayed Pool Name, so this case should sort by the displayed pair name (and fee if needed) instead of the raw path.

packages/web/src/layouts/earn/components/earn-my-positions/earn-my-positions-header/EarnMyPositionsHeader.tsx
```diff
@@ -52,8 +51,8 @@ const EarnMyPositionsHeader: React.FC<EarnMyPositionsHeaderProps> = ({
   }, [onOpenVideoGuide]);
```
junghoon-vans, line 51: `renderMyPositionTitle` uses both `isMobile` and `handleOpenVideoGuide`, but neither is included in the dependency array. After a viewport resize, the memoized callback can keep the old `isMobile` value and leave the `VideoGuideTrigger` line-height stale, so these dependencies should be added.

junghoon-vans (approved):

PR #915, opened 2026-06-08
fix(pool): truncate earn add token symbols

junghoon-vans:
## Summary
- Keep the existing token display cutoff behavior: symbols longer than 9 characters are truncated to 9 characters plus plain `...`.
- Apply `displaySymbol` across `/earn/add` pool info, starting/current price, token switch labels, min/max price, deposit ratio, and add-liquidity broadcast messages.
- Preserve raw token symbols for selection values, comparisons, token identity, logo lookup, and transaction logic.

## Tests
- `yarn workspace @gnoswap-labs/web test:ci --runInBand packages/web/src/utils/token-utils.spec.ts packages/web/src/components/common/select-tab/SelectTab.spec.tsx`
- `yarn workspace @gnoswap-labs/web test:ci --runInBand`
- `yarn build`
- `yarn workspace @gnoswap-labs/web lint --file ...` (changed files only; warnings only)

## Notes
- Full `yarn workspace @gnoswap-labs/web lint` is currently blocked by existing unrelated lint errors in `src/hooks/common/use-click-outside.tsx` and `src/hooks/common/use-delay-loading.tsx`.

### DIFF
--- a/packages/web/src/components/common/select-tab/SelectTab.tsx
+++ b/packages/web/src/components/common/select-tab/SelectTab.tsx
@@ -2,13 +2,13 @@ import { SelectButton, SelectTabWrapper } from "./SelectTab.styles";
 import { cx } from "@emotion/css";
 import { useTranslation } from "react-i18next";
 export interface SelectTabItem {
-  value: string;
   display: string;
+  key: string;
 }
 
 interface SelectTabProps {
   selectType: string;
-  list: Array<string | SelectTabItem>;
+  list: string[] | SelectTabItem[];
   onClick: (type: string) => void;
   buttonClassName?: string;
 }
@@ -19,7 +19,7 @@ const SelectTab: React.FC<SelectTabProps> = ({ selectType, list, onClick, button
   return (
     <SelectTabWrapper className="select-tab-wrapper">
       {list.map((item, idx) => {
-        const value = typeof item === "string" ? item : item.value;
+        const value = typeof item === "string" ? item : item.key;
         const display = typeof item === "string" ? t(item) : item.display;
 
         return (
--- a/packages/web/src/components/common/select-tab/SelectTab.tsx
+++ b/packages/web/src/components/common/select-tab/SelectTab.tsx
@@ -1,9 +1,14 @@
 import { SelectButton, SelectTabWrapper } from "./SelectTab.styles";
 import { cx } from "@emotion/css";
 import { useTranslation } from "react-i18next";
+export interface SelectTabItem {
+  value: string;
+  display: string;
+}
+
 interface SelectTabProps {
   selectType: string;
-  list: string[];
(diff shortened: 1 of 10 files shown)


packages/web/src/components/common/select-tab/SelectTab.tsx
```diff
@@ -1,9 +1,14 @@
...
 import { useTranslation } from "react-i18next";
+export interface SelectTabItem {
+  value: string;
+  display: string;
+}
+
 interface SelectTabProps {
   selectType: string;
-  list: string[];
+  list: Array<string | SelectTabItem>;
```
notJoon: `Array<string | SelectTabItem>` allows a mixture of string and `SelectTabItem`s in the same array. if that isn't an intended use case, `string[] | SelectTabltem[]` might be a better fit since it prevents invalid mixed states and makes the type contract clearer

notJoon: Not a blocker but we already have a `(value, label)` separation pattern in `components/common/select-tab-v2/SelectTabV2.tsx`. it may be worth aligning the field names in `SelectTabItem` with the key structure used in `SelectTabV2`, or deprecating `SelectTab` and consolidating it into `SelectTabV2`.

packages/web/src/layouts/pool/pool-add/components/additional-info/exchange-rate-graph/pair-ratio/PairRatio.tsx
```diff
@@ -65,7 +65,7 @@ export function PairRatio({
     <PairRatioWrapper>
       {!loading && (
         <MissingLogo
-          symbol={displayTokenSymbol}
+          symbol={!isSwap ? pool.tokenA?.symbol : pool.tokenB?.symbol}
```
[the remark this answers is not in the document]
junghoon-vans: Thanks, fixed in `e654162d`. `MissingLogo.symbol` now applies the same `replaceGnotSymbol()` transformation as the displayed ratio text, so the WGNOT fallback label stays consistent with `GNOT`.

packages/web/src/components/common/select-tab/SelectTab.tsx
junghoon-vans: Good point. I aligned `SelectTabItem` with the `SelectTabV2` key structure by using `{ display, key }` in `e654162d`. Full `SelectTab`/`SelectTabV2` consolidation sounds like a separate cleanup from this token-symbol fix.

junghoon-vans: Updated in `e654162d`. `SelectTab` now uses `list: string[] | SelectTabItem[]`, so callers cannot pass a mixed string/object list.

notJoon (approved):
lgtm 👍

PR #931, opened 2026-07-02
fix: align frontend repositories with API contract

notJoon:
## Description

This PR fixes frontend/API contract mismatches for account balances, user positions, dashboard account activity, and position response mapping.

The changes keep the frontend request paths relative to the configured API base URL, unwrap API gateway responses consistently, and lock the expected behavior with focused regression tests.

### Account Balances

Current API contract:

```http
GET /v1/users/{address}/balances
```

The frontend `AxiosClient` receives an API base URL that can already include `/v1`, for example:

```env
NEXT_PUBLIC_DEFAULT_CHAIN_API_URL="https://dev.api.gnoswap.io/v1"
```

Repository calls therefore use a path relative to that base URL:

```ts
/users/${address}/balances
```

Current response shape:

```json
{
  "message": "Success",
  "data": [
    {
      "path": "gno.land/r/demo/token",
      "amount": "100"
    }
  ]
}
```

Fixes applied:

- Changed `AccountRepositoryImpl.getBalances` from `/v1/users/${address}/balances` to `/users/${address}/balances`.
- Changed parsing from `response.data.balances` to `response.data.data`.
- Updated `AccountBalancesResponse` to `{ message, data }`.
- Updated `AccountBalanceModel` to the current item shape `{ path, amount }`.
- Updated account mock balance JSON to use `{ path, amount }`.
- Added regression tests for the runtime repository response and mock repository shape.

### User Position List Query

Current API contract:

```http
GET /v1/users/{address}/position?poolPath=...&page=...&limit=...&withClosed=...&withAvailableStake=...
```

Supported query parameters:

```text
poolPath
page
limit
withClosed
withAvailableStake
```

There is no backend `closed` or `isClosed` query parameter.

Fixes applied:

- Removed `isClosed` from the position repository request options.
- Stopped sending the legacy `closed=` query parameter.
- Kept `isClosed` only as a React Query client-side filter so existing UI behavior is preserved.
- Removed `isClosed` from the repository call in `useGetPositionsByAddress`.
- Added `withClosed: false` to stake, add liquidity, and remove liquidity position queries that need open positions.
- Added a regression test that verifies `withClosed` is sent and `closed` is not sent.

### Dashboard Account Activity

Current API contract:

```http
GET /v1/users/{address}/activity
```

Current response shape:

```json
{
  "message": "Success",
  "data": [
    {
      "txHash": "...",
      "actionType": "...",
      "success": true
    }
  ]
}
```

Fixes applied:

- Updated `DashboardRepositoryImpl.getAccountOnchainActivity` to request `{ data: ActivityResponse }`.
- Changed return value from the raw response body to `data.data`.
- Added a regression test for the response unwrap.

### Position `totalDailyRewardsUsd`

Current API contract:

The backend `Position` response does not expose `totalDailyRewardsUsd`.

Fixes applied:

- Made `PositionResponse.totalDailyRewardsUsd` optional.
- Updated `PositionMapper.from` to default a missing value before formatting.
- Added a regression test for missing `totalDailyRewardsUsd`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Closed positions are excluded consistently across pool add, remove, stake, and portfolio position views.
  * Position daily reward totals now safely default when the value is missing.
  * Account balances and on-chain activity now load from the updated backend response format.
* **Tests**
  * Added/updated unit tests to validate closed-position filtering behavior, reward defaulting, and the new balance/activity response handling.

### DIFF
--- a/packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
+++ b/packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
@@ -67,7 +67,6 @@ const AdditionalInfoContainer: React.FC = () => {
   }, [activePoolPath, pools]);
 
   const { totalPositionCount, loading: isLoadingTotalPositionCount } = usePositionData({
-    isClosed: false,
     poolPath: activePoolPath || "",
     withClosed: false,
     limit: 1,
@@ -85,7 +84,6 @@ const AdditionalInfoContainer: React.FC = () => {
   }, [totalPositionCount]);
 
   const { positions, loading: isLoadingPosition } = usePositionData({
-    isClosed: false,
     poolPath: activePoolPath || "",
     withClosed: false,
     limit: positionLimit,
--- a/packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
+++ b/packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
@@ -69,6 +69,7 @@ const AdditionalInfoContainer: React.FC = () => {
   const { totalPositionCount, loading: isLoadingTotalPositionCount } = usePositionData({
     isClosed: false,
     poolPath: activePoolPath || "",
+    withClosed: false,
     limit: 1,
     queryOption: {
       enabled: !!activePoolPath,
@@ -86,6 +87,7 @@ const AdditionalInfoContainer: React.FC = () => {
   const { positions, loading: isLoadingPosition } = usePositionData({
     isClosed: false,
     poolPath: activePoolPath || "",
+    withClosed: false,
     limit: positionLimit,
     queryOption: {
       enabled: !!activePoolPath,
--- a/packages/web/src/repositories/account/account-repository.spec.ts
+++ b/packages/web/src/repositories/account/account-repository.spec.ts
@@ -104,7 +104,7 @@ describe("get balances", () => {
   });
 
   it("returns mock balances with the API item shape", async () => {
-    const balances = await new AccountRepositoryMock(localStorageClient).getBalances(defaultAccountInfo.address);
+    const balances = await new AccountRepositoryMock(localStorageClient as StorageClient<unknown>).getBalances();
 
     expect(balances[0]).toEqual({
       path: "gno.land/r/demo/tong_token",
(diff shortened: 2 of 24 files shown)


packages/web/src/layouts/pool/pool-add/containers/additional-info-container/AdditionalInfoContainer.tsx
```diff
@@ -69,6 +69,7 @@ const AdditionalInfoContainer: React.FC = () => {
   const { totalPositionCount, loading: isLoadingTotalPositionCount } = usePositionData({
     isClosed: false,
```
junghoon-vans: is this required?

notJoon: https://github.com/gnoswap-labs/gnoswap-interface/pull/931/commits/6024579d26d13916bd52ddbd7de0e41dcb4acd24

junghoon-vans (commented):
One issue I noticed in `WalletMyPositionsHeader.tsx`: the header requests positions with `withClosed: isClosed`. On the default view `isClosed` is false, so the API filters closed positions out of the response. That means `positions.some(position => position.closed)` becomes false even for accounts that do have closed positions, and the “show closed” switch never renders, leaving users no way to toggle closed positions on.

Could we determine the switch visibility from an all-positions probe/count comparison, or fetch this header check with `withClosed=true`, while keeping the card list itself controlled by the UI toggle?

packages/web/src/repositories/account/account-repository.spec.ts
```diff
@@ -61,7 +64,7 @@ describe("get account", () => {
       message: "Get account.",
       data: null,
     });
-    let error: any = null;
+    let error: unknown = null;
```
junghoon-vans, line 67: Small type-check nit: after changing this to `unknown`, the later `error?.status` access is no longer narrowed and `tsc --noEmit` reports it. In this test, `expect(error).toMatchObject({ status: 1000 })` would avoid the direct property read while keeping the intent.

notJoon: Applied both findings https://github.com/gnoswap-labs/gnoswap-interface/pull/931/commits/03a7ce836399633515a0a9930252c8765a1702f8

junghoon-vans (dismissed):

junghoon-vans (approved):

PR #943, opened 2026-07-29
refactor: remove token data side effects from price-only views

junghoon-vans:
## Summary
- remove remaining bare/tokens-only/token-prices-only `useTokenData` usage from the follow-up target views
- add a lightweight `useTokenPricing` for governance price lookup without balance side effects
- expand the token-data optimization guard spec to cover the migrated consumers

## Validation
- `yarn jest --testPathPattern="earn-token-data-optimization|GovernanceSummary|WalletPositionCardListContainer"`
- `yarn next lint --file ...` (warnings only: existing hook dependency warnings; spec files are ignored by lint config)
- `yarn workspace @gnoswap-labs/web build`

## Summary by CodeRabbit

* **New Features**
  * Improved token pricing and metadata retrieval across claims, governance, pools, swaps, staking, and portfolio views.
  * Added more reliable USD value calculations with safe handling while token data is loading or unavailable.
  * Improved loading indicators and reward information display during token data retrieval.

* **Bug Fixes**
  * Prevented incomplete token data from causing incorrect values or premature rendering.

### DIFF
--- a/packages/web/src/hooks/token/data/use-token-pricing.tsx
+++ b/packages/web/src/hooks/token/data/use-token-pricing.tsx
@@ -10,19 +10,23 @@ import { useGetAllTokenPrices, useGetTokens } from "@query/token";
  * Use when a consumer needs tokens + tokenPrices + getTokenUSDPrice only.
  */
 export const useTokenPricing = () => {
-  const { data: { tokens = [] } = {} } = useGetTokens();
+  const { data: { tokens = [] } = {}, isFetched: isFetchedTokens } = useGetTokens();
   const { data: tokenPrices = {} } = useGetAllTokenPrices();
 
   const getTokenUSDPrice = useCallback(
     (tokenAId: string, amount: bigint | string | number) => {
       const tokenUSDPrice = tokenPrices[tokenAId]?.usd || "0";
-      if (!tokenUSDPrice || Number.isNaN(amount)) {
+      const amountValue = BigNumber(amount.toString());
+      const priceValue = BigNumber(tokenUSDPrice);
+      const usdValue = amountValue.multipliedBy(priceValue);
+
+      if (!amountValue.isFinite() || !priceValue.isFinite() || !usdValue.isFinite()) {
         return null;
       }
-      return BigNumber(amount.toString()).multipliedBy(tokenUSDPrice).toNumber();
+      return usdValue.toNumber();
     },
     [tokenPrices],
   );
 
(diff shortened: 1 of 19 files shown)


packages/web/src/hooks/token/data/use-token-pricing.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+      if (!amountValue.isFinite() || !priceValue.isFinite() || !usdValue.isFinite()) {
+        return null;
+      }
+      return usdValue.toNumber();
+    },
+    [tokenPrices],
+  );
+
+  return { tokens, tokenPrices, getTokenUSDPrice, isFetchedTokens };
```
notJoon: `tokenPrices` is returned but unused by the caller, might be worth removing it

packages/web/src/hooks/token/data/use-token-pricing.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useGetAllTokenPrices, useGetTokens } from "@query/token";
+
+/**
+ * Lightweight pricing hook that provides token metadata and USD pricing without
+ * balance side-effects (no useGetGrc20Balances, no updateBalances, no RPC calls).
+ *
+ * Use when a consumer needs tokens + tokenPrices + getTokenUSDPrice only.
+ */
+export const useTokenPricing = () => {
+  const { data: { tokens = [] } = {}, isFetched: isFetchedTokens } = useGetTokens();
```
notJoon, line 13: `TokenListResponse` currently only wraps the `tokens` array, which results in repeated nested default value handling at each call site.
https://github.com/gnoswap-labs/gnoswap-interface/blob/f745ecbf90b3d6bb7e4298e97ec921ad354d83fa/packages/web/src/repositories/token/response/token-list-response.ts#L3-L5

returning the array directly from `useGetTokens` would simplify both the interface and all of its consumers.

but this feels slightly out of scope for th current changes so it should be ok to address it as a follow up

notJoon (commented):
left some minor suggestions. otherwise lgtm 👍

notJoon (dismissed):

packages/web/src/hooks/token/data/use-token-pricing.tsx
junghoon-vans: Agreed. This would be a broader `useGetTokens` API shape cleanup, so let's handle it in a separate follow-up PR rather than expanding the scope of this one.

notJoon (approved):

PR #969, opened 2026-09-01
refactor(tx): send GRC20 approve and transfer as MsgRun

jinoosss:
## Summary

GRC20 balance mutations were sent as `MsgCall` into the gnoswap `common` realm (`common.Approve`, `common.Transfer`). This PR moves them to `MsgRun`, running an ephemeral `main` package that calls the GRC20 registry directly — the approach [Adena uses for GRC20 transfers](https://github.com/onbloc/adena-wallet/blob/3e91e597a966ee0ab14bd45a9bcd67a2717186c2/packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx#L328-L387).

Consecutive approves are batched into a **single** `MsgRun` message instead of one message per token/spender pair.

## Generated source

A swap that approves two tokens now produces one message carrying:

```gno
package main

import (
	grc20reg "gno.land/r/demo/defi/grc20reg"
)

func main(cur realm) {
	grc20reg.Approve(0, cur, "gno.land/r/gnoswap/v1/gns", address("g1pool..."), 1250000)
	grc20reg.Approve(0, cur, "gno.land/r/onbloc/bar", address("g1pool..."), 3000000)
}
```

## Why MsgRun, and why the actor is unchanged

The registry helpers deployed on chain are:

```go
func Approve(_ int, rlm realm, tokenKey string, spender address, amount int64) {
	checkErr(MustGet(tokenKey).RealmTeller(0, rlm).Approve(0, rlm, spender, amount))
}
```

The caller realm arrives as a **secondary** parameter, and `RealmTeller(0, rlm)` makes that realm the actor. As the registry's own doc comment puts it, these are realm-only by construction: `MsgCall` cannot build a realm argument, so a signing user cannot reach them at all — `MsgRun` is the only path.

`main` therefore forwards its own `cur` non-crossing (`0, cur`) rather than crossing into a wrapper. Under `MsgRun` that realm is the reserved `gno.land/e/<caller>/run`, and gnovm's `DerivePkgBech32Addr` returns the caller address itself for that path (`IsGnoRunPath` short-circuit). The balance and allowance owner is the signer, exactly as it was under `MsgCall` into `common`.

The helpers panic on failure via `checkErr`, so the generated statements need no error handling of their own.

## Changes

- **New** `transaction-messages/run.ts` — `makeGRC20ApproveRunMessage` (accepts a list of approves, emits one package) and `makeGRC20TransferRunMessage`, plus `gnoStringLiteral` / `gnoInt64Literal`. Token paths and addresses are rendered as escaped Go string literals and amounts are validated as non-negative integers, so no caller-supplied value can inject statements into the generated source.
- `makeTokenApproveMessages` groups a consecutive approve block by caller and emits one `MsgRun` per group. Approve resets collapse the same way.
- `makeTransferGRC20TokenMessage` / `makeTransferGRC20TokenMessages` switched over; the duplicated inline transfer message in `wallet.message.ts` was removed in favour of the shared builder.
- Run messages carried through the whole pipeline: protocol union + `isRunMessage` guard, `transformMessages`, the Adena client, the social wallet client (`makeMsgRunMessage`), and the transaction service.
- `TransactionMessage` in `transaction-messages/common.ts` is now a union of `TransactionCallMessage | TransactionRunMessage`, with `isTransactionCallMessage` / `isTransactionRunMessage` guards.
- New `PACKAGE_GRC20_REGISTRY_PATH` constant (`NEXT_PUBLIC_PACKAGE_GRC20_REGISTRY_PATH`, defaulting to `gno.land/r/demo/defi/grc20reg`) supplies the generated import path.

### Bug fix included

`encodeMessageValue` defaulted `MsgRun.send` to `"0ugnot"`. A zero-amount coin fails `MsgRun.ValidateBasic` (`Coins.validate` rejects non-positive amounts), so "no send" has to be an empty string. This branch was dead code until now, but it is on the social-wallet signing and gas-estimation path once run messages exist.

## Out of scope

- **NFT approve** (`gnft`, `PACKAGE_NFT_PATH`) stays `MsgCall` — GRC721, not a GRC20 balance transfer.
- **wugnot `Deposit` / `Withdraw`** stay `MsgCall` — native coin send into the wugnot realm.

## Reviewer notes

- `PACKAGE_COMMON_PATH` is no longer referenced by any source file. The constant and its env var were kept so deploy configs stay valid; say the word if you'd rather drop them.
- `MsgRun` compiles a package per message, so it costs more gas than the equivalent `MsgCall`. Batching approves offsets part of this, but `DEFAULT_GAS_WANTED` is worth re-measuring on dev before release.
- The social wallet approval modal shows these entries as function `Run` with no `pkg_path`; the full source is visible in the raw message viewer.
- `run.spec.ts` pins the generated gno source verbatim and covers the escaping/validation guards. Other specs compare against `run.test-fixtures.ts`, which rebuilds the expected package independently of the production template.

## Test plan

- [x] `yarn test:ci` — 160 suites / 732 tests pass
- [x] eslint clean on all changed files
- [x] Registry signatures verified against `dev.gnoswap` (`vm/qfile` on `gno.land/r/demo/defi/grc20reg`)
- [x] Manual: swap, add/remove liquidity, stake, delegate, launchpad deposit and wallet GRC20 send against dev chain

## Summary by CodeRabbit

* **New Features**
  * Added support for run-based transactions across wallet integrations and transaction processing.
  * Token transfers, approvals, and approval resets now use the GRC20 registry transaction format.
  * Consecutive token approvals can be combined into a single transaction.
  * Added configuration for the GRC20 registry package.

* **Bug Fixes**
  * Prevented invalid or oversized token amounts from generating transactions.
  * Added clearer error handling when run transactions cannot be created.

* **Tests**
  * Expanded coverage for run transactions, token operations, amount validation, and message batching.


packages/web/src/common/clients/wallet-client/transaction-messages/run.ts
```diff
@@ -0,0 +1,166 @@
...
+ * integer is rejected instead of being written into the source.
+ */
+export function gnoInt64Literal(amount: string | number | bigint): string {
+  const amountBN = BigNumber(amount.toString());
+
+  if (!amountBN.isFinite() || !amountBN.isInteger() || amountBN.isNegative()) {
+    throw new TransactionMessageError("FAILED_BUILD_RUN_MESSAGE", amount);
+  }
+
+  return amountBN.toFixed(0);
```
junghoon-vans, line 82: `gnoInt64Literal` only checks that the value is a non-negative integer, but it does not enforce `<= 9223372036854775807`. For example, `9223372036854775808` is emitted verbatim and then fails when Gno converts the literal to the `int64` amount expected by `grc20reg.Approve`/`Transfer`. Please reject values above `MAX_INT64_STR` here and add an upper-bound test.

junghoon-vans (approved):

PR #972, opened 2026-09-09
refactor: model transaction modal state as a discriminated union

notJoon:
## Description

Transaction modal fields were independently nullable, allowing invalid combinations such as a successful transaction with no hash. The state, update API, and modal props now use a discriminated union, so TypeScript requires the fields each status needs and narrows them when rendering.

Before:

```ts
export const transactionModalData = atom<{
  status: "loading" | "rejected" | "success" | "error";
  title: string | null;
  description: string | null;
  txHash: string | null;
  callback?: () => void;
} | null>(null);
```

After:

```ts
type TransactionModal = (
  | { status: "loading"; description: string }
  | { status: "rejected" }
  | { status: "error"; title: string; description: string }
  | { status: "success"; title: string; description: string; txHash: string }
) & { callback?: () => void };

export const transactionModalData = atom<TransactionModal | null>(null);
```

Reposition transactions now retain their response hash instead of overwriting it with `null`, keeping the explorer link usable. The shared success handler falls back to the error state when a response has no hash. Loading descriptions and close callbacks are preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction confirmation handling across loading, success, failure, and rejection states.
  * Missing transaction hashes now display a clear default error message instead of leaving the confirmation flow incomplete.
  * Transaction success and error details are displayed more consistently, including scanner links where available.
  * Repositioning liquidity now reliably shows completion details and returns to the previous screen after confirmation.


packages/web/src/hooks/common/use-broadcast-handler.ts
```diff
@@ -55,51 +55,51 @@ export const useBroadcastHandler = () => {
...
 
   const broadcastSuccess = useCallback(
-    (content?: SnackbarContent, callback?: () => void) => {
+    (content: SnackbarContent, callback?: () => void) => {
+      if (!content.txHash) {
+        setTransactionModalData({
+          status: "error",
+          title: BROADCAST_ERROR_VALUE.DEFAULT.title,
+          description: BROADCAST_ERROR_VALUE.DEFAULT.description,
+          callback,
```
aronpark1007: Not a blocker, but this preserves the success callback even after downgrading to an error state. Since closing the error modal still runs `transactionModalData.callback()`, a missing-hash success response could show a failure modal and then still navigate back. Consider dropping the callback here or only running it for `status === "success"`.

notJoon: seems legit. I removed the callback for the missing tx hash case and updated the test to ensure it isn’t fired after dismissing the error modal.

https://github.com/gnoswap-labs/gnoswap-interface/pull/972/commits/668095d7ae2981ee8579fa183c2c4476f38cb8e6

aronpark1007 (approved):

PR #986, opened 2026-09-15
refactor: remove gas simulation from transaction confirm flows

jinoosss:
## Summary

Removes the gas simulation that ran before broadcasting transactions and the live gas estimate shown on the swap card. Repositories now fall back to their default `gasFee` / `gasWanted` values. The underlying `transactionGasService.estimateGas` and related utilities are kept; they are simply no longer called.

### Swap
- `useSwap` no longer builds a simulation document on every input change or runs `useNetworkFee`.
- Swap / wrap / unwrap requests no longer attach simulated `gasFee` / `gasUsed`.
- Removed the gas fee UI: gas icon and amount in the swap card header, the Gas Fee row in the detail panel, the Gas Fee row in Confirm Swap modal, and the Gas Fee item in the button tooltip.
- Dropped `gasFee`, `gasFeeUSD`, `gasEstimateSuccess` from `SwapSummaryInfo`; updated stories and specs.

### Pre-broadcast simulation (social wallet path)
Removed the `createDocument -> createTransaction -> estimateNetworkFee` block that ran before every social wallet transaction in:
- stake / unstake / remove / increase / decrease position
- reposition (remove, swap remain token, reposition)
- claim / claim all
- create pool / add liquidity
- create / remove / collect external incentive
- GNOT / GRC20 transfer

Where the social and Adena builders became identical after the removal, they were collapsed into a single action. The reposition flow keeps both builders since they differ in broadcast handling.

### Cleanup
- Deleted `useNetworkFee`, `useGetEstimateGasInfo`, `buildGasInfo` (+ spec), `hooks/gas/types.ts`, and `QUERY_KEY.gasInfo`.
- Removed the now-unused `rpcProvider` parameters from affected hooks and their callers.

## Not changed
- `TransactionApprovalModal` still shows the Network Fee row; it displays the fixed fee from the document, not a simulation.
- `getRepositionAmountsWithSwapSimulation` is a price simulation and is untouched.

## Verification
- `tsc --noEmit`: no new errors (only pre-existing spec / svg module errors remain).
- ESLint and Prettier pass on all changed files.
- Jest suites under `components/swap`, `layouts/swap`, `hooks/gas`, `utils/swap-utils`: 8 suites / 52 tests passed.


junghoon-vans (dismissed):
Looks good overall. Two non-blocking cleanup notes:

- `rpcProvider` is still required by `createPool`, `addLiquidity`, `claimAll`, and `claim` even though these paths no longer use it. Please consider removing the parameters and remaining `GnoProvider` imports.
- `SwapRouteInfo.gasFee` / `gasFeeUSD` and their computation in `useSwapHandler` are now unused by the route UI. Please consider removing this dead data and the related story fixtures.

notJoon (dismissed):

junghoon-vans (approved):

notJoon (approved):
LGTM 👍

PR #987, opened 2026-09-15
chore(deps): upgrade gno-js-client and tm2-js-client to v3

jinoosss:
## Summary

Upgrades the Gno JS clients to the latest published versions and pins them to exact releases.

| Package | Before | After |
| --- | --- | --- |
| `@gnolang/gno-js-client` | `^1.4.2` | `3.0.0` |
| `@gnolang/tm2-js-client` | `^1.3.1` | `3.0.0` |
| `@gnolang/tm2-rpc` | – | `1.0.0` (new, required to build a provider) |

> Note: the latest `@gnolang/tm2-js-client` is `3.0.0`, not `2.2.0` — `2.2.0` is the latest `2.x` of `gno-js-client`. `gno-js-client@3.0.0` itself requires `@gnolang/tm2-js-client@^3.0.0`, so both move to `3.0.0`.

Both releases are breaking, so this PR also migrates the call sites.

### 1. Provider construction is now asynchronous

`GnoJSONRPCProvider` is built on top of a `Tm2Client` and exposes a static `create()` factory instead of a URL constructor. `GnoProvider` follows the same shape, and `GnoswapServiceProvider` awaits it (ignoring the result if the effect has been superseded).

`getGasPrice` no longer needs the hand-rolled `fetch` against `baseURL` — it goes through `this.client.abciQuery`, which made `src/utils/gno-utils.ts` dead code.

### 2. `Long` → `bigint`

`TxFee.gas_wanted` and the return value of `estimateGas` are `bigint` now:

- `documentToTx` / `documentToDefaultTx` build the fee with `BigInt(...)`
- `TransactionGasServiceImpl.estimateGas` converts back to `number` at the boundary
- `TxSignature` gained a required `session_addr` field

### 3. Social wallet signing

`@adena-wallet/sdk` still bundles `tm2-js-client` v1, whose encoder calls `gas_wanted.equals()` / `.toNumber()` — both of which throw on a `bigint`. `SocialWalletClient.sign` now rebuilds the document through the SDK's own `TransactionBuilder`, so the fee stays in the shape the SDK expects. The app keeps using the v3 `Tx` everywhere else.

### 4. Message types moved to the package root

`MsgRun` is exported from `@gnolang/gno-js-client` directly; the `bin/proto/gno/vm` deep import no longer exists. `MsgCall` / `MsgRun` / `MsgAddPackage` also gained a `max_deposit` field, reflected in the test fixtures.

### 5. `@scure/base` pinned for `@cosmjs/encoding`

`tm2-js-client` v3 pulls in `@cosmjs/encoding@0.38.1`, which calls `bech32.decode(address, Infinity)`. `@scure/base` 2.3.0 added a `anumber(limit)` guard that rejects `Infinity`, so any `getStatus()` call threw `RangeError: limit: expected safe integer, got Infinity`. A scoped resolution keeps that dependency on `2.2.0`.

### 6. Node 18 can no longer run this dependency tree

`uuid@14`, `@scure/base@2` and `@cosmjs/*` are published as ESM only, while the CommonJS entrypoints of `tm2-js-client` and `@cosmjs/encoding` `require()` them. That only works from Node 20.19, where `require(esm)` is enabled by default, so `next build` failed on Node 18 while collecting page data:

```
Error: require() of ES Module .../@scure/base/index.js
from .../@cosmjs/encoding/build/bech32.js not supported.
code: 'ERR_REQUIRE_ESM'
```

`transpilePackages` was tried first, but it needs an ever-growing allowlist of transitive packages and still fell through to another externalized ESM `require()`. CI (`node-version: 18`) and the Dockerfile (`node:18-alpine`) were also the only places still on 18, while `.nvmrc` already pinned `20.19.0`, so they now follow `.nvmrc`.

### 7. Jest has to transform ESM-only dependencies

`uuid@14`, `@scure/base@2`, `@noble/hashes@2` and `@cosmjs/*` are published as ESM only, and `next/jest` hard-codes `node_modules` as non-transformed while only allowing configs to *append* to `transformIgnorePatterns`. The resolved config is patched after the fact so those packages go through the transformer. Enclosing scopes have to be listed too, otherwise the pattern still matches on the outer `node_modules/` segment of a nested dependency.

## Test plan

- [x] `yarn workspace @gnoswap-labs/web tsc --noEmit` — no new errors (remaining ones are pre-existing)
- [x] `yarn test:ci` on Node 20.19.0 — 164 suites / 744 tests pass (20 suites were failing before the Jest change)
- [x] `yarn build` on Node 20.19.0 — production build succeeds, including the `Collecting page data` step that failed in CI
- [x] Confirmed the same build fails on Node 18 with `ERR_REQUIRE_ESM`, and that `require()` of `@gnolang/tm2-js-client` / `@cosmjs/encoding` succeeds on 20.19.0 but not on 18
- [x] `yarn lint` — warnings only, unchanged from before
- [x] Live smoke test against `https://dev.rpc.gnoswap.io`: `getStatus`, `getBlockNumber`, `getBalance`, `getGasPrice` and `evaluateExpression` all return the expected values

## Risks

- Provider creation now performs a network round trip and can fail. If the RPC endpoint is unreachable, `rpcProvider` stays `null` and the app never reports itself as initialized — previously the constructor could not fail. Worth watching on a flaky endpoint.
- Transaction signing paths (Adena extension and social wallet) are not covered by automated tests and should be verified manually before merging.
- The runtime floor moves to Node 20.19 — anything deploying this image or running the app outside CI has to be on that version or newer. Node 18 reached end of life in April 2025.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved wallet transaction signing compatibility for sends, package deployments, contract calls, and contract runs.
  - Updated transaction fee and gas handling to support large gas values more reliably.
  - Improved network connection handling, including safer recovery when connections change or fail.
  - Added a retry option when the RPC connection cannot be established.
  - Updated transaction message formats for compatibility with the latest Gno client libraries.

- **Testing**
  - Updated automated tests to reflect current transaction message and deposit formats.


packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
```diff
@@ -145,7 +145,26 @@ const GnoswapServiceProvider: React.FC<React.PropsWithChildren> = ({ children })
       }),
     );
     setRouterApiClient(new AxiosClient(network.routerUrl));
-    setRPCProvider(new GnoProvider(network.rpcUrl || ""));
+
+    // Creating a provider requires a round trip to the node since gno-js-client v3,
+    // so ignore the result once the effect has been superseded.
+    let stale = false;
+    GnoProvider.create(network.rpcUrl || "")
```
junghoon-vans, line 167: Please invalidate the current RPC provider before switching the other network clients. The API and router clients are replaced synchronously, while the old non-null `rpcProvider` remains available until this promise resolves. Since `loadedProviders` therefore stays true, descendants can issue requests with new-chain API clients and the previous chain's RPC provider during the transition. Clear the provider before swapping clients, or publish a single chain-keyed provider bundle atomically after creation completes.

packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
```diff
@@ -145,7 +145,26 @@ const GnoswapServiceProvider: React.FC<React.PropsWithChildren> = ({ children })
...
+      .then(provider => {
+        if (!stale) {
+          setRPCProvider(provider);
+        }
+      })
+      .catch(error => {
```
junghoon-vans, line 181: Please handle initialization failure before keeping children behind `loadedProviders`. If this initial connection rejects, `rpcProvider` remains null and the render gate suppresses the entire app tree, including the error page; there is also no retry. `Tm2Client.connect` has no finite HTTP deadline, so a blackholed endpoint can leave the UI blank indefinitely without reaching this catch. Render a retryable failure state outside the gate and bound or cancel the connection attempt.

junghoon-vans (changes requested):
Requesting changes for two regressions introduced by asynchronous RPC provider initialization.

junghoon-vans:
Non-blocking existing issue: the reconstruction logic in `encodeMessageValue` was already present on `develop`, but it does not preserve all VM message fields. It omits `MsgAddPackage.send` and `MsgAddPackage.max_deposit`, and forces `MsgCall.max_deposit` and `MsgRun.max_deposit` to an empty string. A local encode/decode reproduction with `max_deposit: "123ugnot"` decoded to an empty value.

Because the reconstruction logic predates this PR, I am not classifying it as an upgrade regression. However, this v3 migration should ideally preserve every message field and add round-trip coverage for non-empty `send` and `max_deposit` values.

packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
jinoosss: Fixed in 7d46cbbfe by publishing the three clients as one chain-keyed bundle.

`gnoswapApiClient`, `routerApiClient` and `rpcProvider` are no longer separate states. They are held in a single `ChainClients` value that is only written once `GnoProvider.create` resolves, so the previous chain's consistent set stays in place for the whole transition instead of the API clients moving ahead of the RPC provider.

I kept the previous set on a failed switch rather than clearing it, so a failed chain switch leaves the app usable on the chain it was already on.

The `chainId` on the bundle is also used to skip the effect when the chain has not changed. Connecting is a real network round trip now, and the effect re-runs when `loadedProviders` flips, so without that guard every load opened two connections.

packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
jinoosss: Fixed in 7d46cbbfe, both parts.

Bounded connection: `GnoProvider.create` now races `Tm2Client.connect` against a 15s deadline. To be precise about what that does and does not give us — `HttpEndpoint` only exposes `url` and `headers`, and the client passes no `AbortSignal`, so the in-flight request is not actually cancelled. What is guaranteed is that the caller always settles and reaches the error path instead of staying pending.

Retryable failure state: the provider now tracks `connectionFailed` and renders `RpcConnectionFailed` with a retry button when nothing has loaded yet. It renders in place of the gate rather than behind it, so a failed first connection no longer produces a blank tree. It cannot reuse `Custom500`, since that pulls in `HeaderContainer`, which needs the very context that is unavailable at that point, so it is a small self-contained component with inline styles. Retry bumps a counter in the effect dependencies.

Note it is shown only when there is nothing loaded. A failed chain switch keeps the previous working clients instead, as described in the other thread.

packages/web/src/common/clients/gno-provider/gno-provider.ts
```diff
@@ -1,20 +1,30 @@
...
 export class GnoProvider extends GnoJSONRPCProvider {
-  constructor(baseURL: string) {
-    super(baseURL);
+  /**
+   * Connects to the node and creates a provider bound to it.
+   * Since v3 the provider is built on top of a Tm2Client, so instantiation
+   * requires a round trip to the node and can no longer be done synchronously.
+   */
+  public static async create(baseURL: string): Promise<GnoProvider> {
+    return new GnoProvider(await Tm2Client.connect(baseURL));
```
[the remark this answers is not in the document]
jinoosss: Addressed in 7d46cbbfe. `GnoProvider.create` now races the connection against a 15s deadline.

Confirmed your reading of the transport: `HttpEndpoint` in `@gnolang/tm2-rpc@1.0.0` only carries `url` and `headers`, and there is no `AbortSignal` plumbing, so the request cannot actually be cancelled. The timeout bounds the wait and guarantees the caller settles; the comment on the method says so explicitly rather than claiming cancellation.

packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
```diff
@@ -145,7 +145,26 @@ const GnoswapServiceProvider: React.FC<React.PropsWithChildren> = ({ children })
...
+        console.error("Failed to connect to the RPC provider", error);
+        if (!stale) {
+          setRPCProvider(null);
```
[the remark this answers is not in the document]
jinoosss: Addressed in 7d46cbbfe. `GnoswapServiceProvider` now tracks the failure and renders a retryable `RpcConnectionFailed` in place of the gate, so a rejected connection no longer suppresses the whole tree. `Custom500` is not reusable here because it renders `HeaderContainer`, which consumes the context that is missing in exactly this situation.

packages/web/src/common/clients/wallet-client/social/social-wallet-client.ts
```diff
@@ -36,7 +38,38 @@ import { getSocialWalletConfig } from "./config";
...
+ * protobufjs `Long` rather than the `bigint` used from v3 on. Going through the
+ * SDK's own builder keeps the fee in the shape its encoder expects.
+ */
+const documentToSDKTx = (document: Document) => {
+  const [gasFee] = document.fee.amount;
+
+  return TransactionBuilder.create()
+    .messages(...document.msgs.map(toSDKMessage))
+    .gasWanted(Number(document.fee.gas) || DEFAULT_GAS_WANTED)
+    .fee(Number(gasFee?.amount) || 0, gasFee?.denom || GNOT_UNIT_DENOM)
```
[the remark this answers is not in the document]
jinoosss: Not applying this one — the overflow is not reachable on these inputs, and I would rather not add a guard that suggests it is.

Both values are produced as JS numbers and only stringified on the way into the document: `createDocument` writes `gas: Math.ceil(args.gasWanted * 1.3).toString()` and `amount: String(args.gasFee)`. `Number()` on those is an exact round trip. The approval modal is the only place a document is edited between creation and signing, and it only changes `memo` (`memoChangeHandler`); gas and fee are display-only there.

The pre-existing `sendTransaction` path in the same file already passes these as numbers to the same builder methods, so a guard here would cover one of two call sites for a case neither can reach. If the builder ever accepts an exact integer type, both should move over together.

jinoosss:
Follow-up on the `encodeMessageValue` thread (the non-blocking one about VM message fields).

Fixed in 7d46cbbfe, with one extra step that matters for on-chain behaviour.

`encodeMessageValue` now copies `send` and `max_deposit` for `MsgAddPackage`, and `max_deposit` for `MsgCall` and `MsgRun`, instead of dropping them. Added `transaction-utils.spec.ts` with round trips that encode and decode each message type with non-empty `send` and `max_deposit`.

The extra step: preserving the field alone would have changed what goes out. `createDocument` emits `max_deposit: "0ugnot"` for `MsgCall`, which only ever reached the wire as `""` because the encoder dropped it. The code already warns in the `MsgRun` branch that a zero-amount coin string fails `ValidateBasic`, so simply preserving it would have started sending `"0ugnot"`. I changed the source to `""` as well, so the encoded value is byte-identical to today while the silent drop is gone.

Scope note: `documentToDefaultTx`, `makeEstimateGasTransaction` and `TransactionGasService.estimateGas` have had no callers since #986 removed gas simulation from the confirm flows. `documentToTx` is still reachable through the social wallet `sign()` fallback. They are left in place; removing dead code is not this PR's job, but it is worth knowing the blast radius here was small.

packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
```diff
@@ -268,7 +312,7 @@ const GnoswapServiceProvider: React.FC<React.PropsWithChildren> = ({ children })
         faucetService,
       }}
     >
-      {loadedProviders && children}
+      {loadedProviders ? children : connectionFailed ? <RpcConnectionFailed onRetry={retryConnection} /> : null}
```
junghoon-vans, line 320: Please also gate rendering on the published client bundle matching the wallet's requested chain. During a chain switch, `walletAccount` and `eventStore` move to the new chain immediately, but `chainClients` intentionally keeps the old API/RPC bundle while the new connection is pending. Since `loadedProviders` remains true, children continue rendering with clients from different chains. If the new connection fails, `connectionFailed` becomes true but this ternary still selects `children`, so the retry UI is unreachable and the mismatch persists indefinitely. Keep children gated until `chainClients.chainId` matches the requested chain and render a retryable switch failure, or revert the wallet network on failure.

junghoon-vans (changes requested):
The initial-load failure handling and VM message field preservation are fixed. One chain-switch blocker remains.

packages/web/src/providers/gnoswap-service-provider/GnoswapServiceProvider.tsx
jinoosss: Good catch, and correct on both halves — the failed-switch case in particular was unreachable-retry, which is worse than what it replaced. Fixed in 255ce197b.

`loadedProviders` is now `chainClients?.chainId === network.chainId` rather than `chainClients !== null`, so children render only while the published bundle belongs to the chain the wallet is asking for. A failed switch therefore falls through to the same retry screen as a failed first connection, and `connectionFailed` is reset when a new attempt starts so a stale flag cannot surface during a pending one.

I also resolved the network lookup once in a `useMemo` and shared it with the effect and `eventStore`; the same `SUPPORT_CHAIN_IDS`/`NetworkData.find` block was duplicated in both, which is how the requested chain and the published chain could drift apart in the first place.

One consequence worth flagging rather than hiding: children now unmount for the duration of a chain switch, so there is a blank window until the new connection resolves (bounded by the 15s deadline). That matches what the initial load already does, and I preferred it over rendering mismatched clients. If you want a switching indicator there instead of nothing, say so and I will add one — I did not want to invent UI inside this PR.

packages/web/src/services/transaction/transaction-service-impl.ts
```diff
@@ -42,7 +42,8 @@ export class TransactionServiceImpl implements TransactionService {
       if (isContractMessage(message)) {
         return makeMsgCallMessage({
           ...message,
-          max_deposit: "0ugnot",
+          // A zero-amount coin string fails ValidateBasic; "no deposit" is empty.
+          max_deposit: "",
```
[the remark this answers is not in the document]
jinoosss: Fair, and it was worse than described — `TransactionMessageOfContract` and `TransactionMessageOfRun` never declared `max_deposit` at all, so there was no typed way to supply one. Fixed in 255ce197b.

Added `max_deposit?: string` to both message types, and `createDocument` now uses `message.max_deposit ?? ""` instead of a hardcoded `""`. Did the same in the social wallet `sendTransaction` path, which had the identical overwrite.

Deliberately not normalising a zero value. In the previous commit I noted that a zero-amount coin string fails `ValidateBasic`, but that is documented in this repo for `MsgRun.send`, and I have not verified it holds for `max_deposit` — so treating `"0ugnot"` as "absent" would be encoding a guess. What the earlier change did was narrower: `createDocument` used to emit `max_deposit: "0ugnot"` while the encoder dropped it, so `""` was what actually reached the wire. Dropping that literal kept the bytes identical once the encoder stopped discarding the field. Now nothing in the app supplies a value, so the result is still `""`, and a caller that supplies one gets it through intact.

Coverage added in `transaction-service-impl.spec.ts`, through `createDocument` as you asked: supplied `max_deposit` survives on `MsgCall` and `MsgRun`, and defaults to `""` when absent. It runs under `@jest-environment node` — importing the service pulls in `@adena-wallet/sdk`, whose web3auth dependency hashes at module load and rejects jsdom cross-realm `Uint8Array`.

junghoon-vans (approved):
Verified the follow-up fixes. Provider rendering is now gated on the requested chain, failed chain switches expose the retry fallback, and `max_deposit` is preserved through document creation and social-wallet transformation. The focused transaction suites pass, and I found no additional regressions in the updated files.

notJoon (approved):

PR #988, opened 2026-09-15
fix(swap): keep full precision for balances above MAX_SAFE_INTEGER

jinoosss:
## Summary

Temporary fix for the "Max amount exceeds balance" error on swap.

When a wallet holds a GRC20 balance whose raw amount exceeds `Number.MAX_SAFE_INTEGER` (2^53), the frontend converted it with `Number(amount)` and the value was rounded **up**. The Max button then requested more than the on-chain balance and the router rejected the swap.

### Reproduction (dev.gnoswap, USDC -> ATOM)

| | raw amount (6 decimals) |
|---|---|
| on-chain `BalanceOf` | `62667447936264477` |
| frontend `Number(amount)` | `62667447936264480` |

Router error:

```
[GNOSWAP-ROUTER-018] insufficient balance for swap
token=..., required=62667447936264480, available=62667447936264477
```

The formatted balance `62,667,447,936.26448` (trailing zero removed) is why lowering the 5th decimal made the swap succeed.

## Changes

- `token-utils.ts`: add `makeDisplayTokenAmountString`, a BigNumber-based conversion that never goes through a JS number.
- `use-token-data.tsx`: add `displayBalanceStringMap` built from the raw balance strings (GRC20 API response and GNOT balance).
- `use-swap-handler.tsx`: swap card balances are formatted from `displayBalanceStringMap`.
- `SwapCardContent.tsx`, `TokenSwap.tsx`: the Max handler uses `BigNumber(...).toFixed()` instead of `parseFloat(...).toString()`, which also rounds values with more than ~16 significant digits.
- `use-swap.tsx`, swap-router request/message types: the swap amount is passed to the router messages as a string instead of `Number(tokenAmount)`.
- `token-utils-precision.spec.ts`: regression test reproducing the on-chain case (old path exceeds the balance, new path matches it exactly).

## Scope / follow-ups

This is a targeted fix for the swap flow only. `displayBalanceMap` and the `balances` atom are still `number` based and are used by portfolio, add-liquidity, launchpad and token-select screens. Those should be migrated to the string-based map in a follow-up.

The GNOT balance from `getBalanceByKey` is still a `number`; it is safe in practice (ugnot balances are far below 2^53) but could be migrated at the same time.

## Test

- `jest src/utils/token-utils-precision.spec.ts src/utils/token-utils.spec.ts src/hooks/swap src/repositories/swap-router`: 49 passed
- `tsc --noEmit`: no new errors in the changed files (existing spec/svg errors are unrelated)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved exact token amounts for large balances throughout swaps, route estimates, approvals, and transactions.
  * Fixed the Max action to use the full on-chain balance without rounding.
  * Improved handling of zero, invalid, and unavailable swap amounts.
  * Preserved decimal precision when displaying token balances and estimated amounts.

* **Tests**
  * Added regression coverage for large-balance swaps, route requests, transaction arguments, and Max-button autofill.


junghoon-vans (changes requested):
Thanks for isolating the original balance conversion and documenting the on-chain reproduction. The string balance map and removal of `parseFloat` in both swap surfaces are the right direction, but the precision path is still incomplete.

## Required changes

### 1. Keep the amount as a string through swap state and route lookup

`useSwap.updateSwapAmount` still converts the exact Max string back to a JavaScript number with `.toNumber()` (`packages/web/src/hooks/swap/data/use-swap.tsx`, around line 185). `useGetRoutes` then converts it again with `Number(request.tokenAmount)` (`packages/web/src/react-query/router/use-get-routes.ts`, around line 39).

Using the reported balance, the exact display amount `62667447936.264477` becomes route raw amount `62667447936264480`, which is 3 raw units above the on-chain balance `62667447936264477`. The transaction message may now use the exact handler string, but the quote and minimum/maximum amounts are calculated from a different rounded value.

Please keep `swapAmount`, `estimatedLiquidityMax`, `GetRoutesRequest.tokenAmount`, `SwapRouteRequest.tokenAmount`, and `tokenAmountLimit` as strings end-to-end. Use BigNumber for comparisons and slippage arithmetic, and remove the remaining `Number()` / `.toNumber()` conversions from this monetary path.

### 2. Do not use the formatted balance for validation or Max

`use-swap-handler.tsx` still passes `displayBalanceStringMap` through `formatPrice(... greaterThan1Decimals: 6)` before storing it in `SwapTokenInfo.tokenABalance`. That same formatted value is used by balance validation and the Max handlers.

For tokens with more than 6 decimals, this truncates the actual balance: Max leaves dust, and a manually entered exact balance can be rejected against the truncated display value. Keep the exact decimal string in `SwapTokenInfo`; apply display formatting only in `useTokenBalancesDisplay`.

### 3. Add production-path regression coverage

`token-utils-precision.spec.ts` reimplements the desired flow in a local `maxAmountRaw()` helper. It does not render/call `SwapCardContent` or `TokenSwap`, exercise `useSwap` / `useGetRoutes`, or build a swap-router transaction message. The test would still pass if the changed handlers or the remaining Number conversions were reverted.

Please add observable regression coverage for the reported value:

- clicking Max forwards `62667447936.264477` unchanged;
- the route request body contains raw amount `62667447936264477`;
- the transaction message args and approval amount contain `62667447936264477`.

The existing `makeDisplayTokenAmountString` tests can remain as focused utility coverage, but they do not replace wiring coverage.

jinoosss:
Thanks for the review. All three points are addressed in 
the latest commit.

**1. String end-to-end**
- `useSwap`: `swapAmount`, `estimatedLiquidityMax` and `tokenAmountLimit` are now decimal strings. Comparisons and the slippage limit use BigNumber; `.toNumber()` is gone from this path.
- `useGetRoutes` forwards `request.tokenAmount` as-is (validated with BigNumber) instead of `Number(...)`.
- `GetRoutesRequest`, `SwapRouteRequest`/`DrySwapRequest`, `ExactSwapRouteMessageRequest` and `swapConfirmModalState.tokenAmountLimit` accept strings. `use-reposition-handle` passes its remainder as a string (`null` when zero).
- The only remaining `Number()` is `guaranteedAmount.amount` in the summary info, which is display-only (`AmountModel.amount` is `number` and used in 19 places); the transaction uses the string `tokenAmountLimit`.

**2. Exact balance for validation and Max**
`SwapTokenInfo.tokenABalance/tokenBBalance` now hold the exact decimal string from `displayBalanceStringMap`. `formatPrice(... 6)` was removed from that path, so tokens with more than 6 decimals no longer lose dust on Max or get rejected on an exact manual entry. Display formatting stays in `useTokenBalancesDisplay`.

**3. Production-path regression coverage** (all using `62667447936.264477`)
- `SwapCardContent.spec.tsx`: clicking Max calls `changeTokenAAmount("62667447936.264477")`.
- `use-swap.spec.tsx`: `updateSwapAmount` -> `useGetRoutes` receives the same string; `swap()` sends `tokenAmount` as the string with a string `tokenAmountLimit`.
- `use-get-routes.spec.tsx`: the hook passes the string to `swapRouterRepository.getRoutes` unchanged.
- `swap-router-repository-impl.spec.ts`: the route request body contains `amount: "62667447936264477"`.
- `swap-router.message.spec.ts`: exact-in args and the approval amount contain `62667447936264477`.

Full suite: 167 suites / 755 tests passing.

junghoon-vans (changes requested):
Thanks—the latest commit addresses the three swap-path issues from my previous review: the swap hook and route lookup now preserve decimal strings, the exact balance is used for validation/Max while formatting stays display-only, and the reported value is covered through the Max, route-body, and transaction-message paths.

Two precision gaps remain:

## Exact-out slippage must round the maximum input upward

`packages/web/src/hooks/swap/data/use-swap.tsx` currently applies `BigNumber.ROUND_DOWN` to both directions when computing `tokenAmountLimit` (around lines 171–177). Rounding down is correct for the exact-in minimum output, but it understates the exact-out maximum input at an atomic-unit boundary.

For a 6-decimal input token, an estimated input of `0.000001` with 0.5% slippage produces `0.000001005`. The current code rounds that to `0.000001` (1 raw unit); the maximum input must round up to `0.000002` (2 raw units) to preserve the configured tolerance. Otherwise an exact-out swap can still fail after any upward quote movement despite the user's slippage setting.

Please select the rounding mode by direction—down for `EXACT_IN`, up/ceiling for `EXACT_OUT`—and cover the atomic-unit boundary.

## The shared request path still permits and produces imprecise numbers

`GetRoutesRequest`, `SwapRouteRequest`, and `ExactSwapRouteMessageRequest` remain `number | string`, and `use-reposition-handle.tsx` still builds the transaction request with `inputAmount.toNumber()`, `outputAmount.toNumber()`, and native-number slippage multiplication (around lines 522–531). The estimate was changed to stringify a number, but the broadcast path can still lose precision before `makeRawTokenAmount` receives it.

This also leaves the request interfaces unable to enforce the string-only invariant requested in the previous review. Please make the monetary request fields strings and build the reposition amounts/limits with BigNumber string output instead of `.toNumber()`/number multiplication.

jinoosss:
Both remaining gaps are addressed in the latest commit.

**Exact-out slippage rounds up**
Added `calculateSlippageLimitAmount(estimatedAmount, slippage, direction, decimals)` in `swap-utils.ts`. EXACT_IN rounds the minimum output down; EXACT_OUT rounds the maximum input up. `useSwap.tokenAmountLimit` and the reposition swap both use it.
Coverage: `swap-utils.spec.ts` checks `0.000001` with 0.5% -> `"0.000002"` (EXACT_OUT) and `"0"` / `"0.000001"` floors (EXACT_IN), plus the large-balance case. `use-swap.spec.tsx` drives the same boundary through the hook and asserts `sendExactOutSwapRoute` receives `tokenAmountLimit: "0.000002"`.

**String-only request path**
- `GetRoutesRequest.tokenAmount`, `SwapRouteRequest` / `DrySwapRequest` `tokenAmount` and `tokenAmountLimit`, and `ExactSwapRouteMessageRequest` are now `string` only (existing specs updated accordingly).
- `use-reposition-handle.tsx`: `currentAmounts` come from `makeDisplayTokenAmountString`, the remainder is `BigNumber(amount).minus(...)` passed as `toFixed()`, and the transaction request uses `inputAmount.toFixed()` / `outputAmount.toFixed()` with the shared slippage helper instead of `.toNumber()` and native multiplication. This also covers the CodeRabbit note on the remainder calculation.

Full suite: 167 suites / 760 tests passing; no new tsc or eslint errors.

junghoon-vans (changes requested):
The exact-out rounding fix is correct, and the shared request interfaces are now string-only. The atomic-unit boundary coverage also matches the intended behavior.

One part of the reposition precision path is still unresolved: the new `toFixed()` calls serialize values that have already passed through JavaScript numbers upstream.

`getRepositionAmountsByPriceRange` still returns `{ amountA: number; amountB: number }` and converts calculated amounts with `Number(...)` / `.toNumber()` (`packages/web/src/utils/reposition-utils.ts`, around lines 20–21 and 45–80). `getRepositionAmountsWithSwapSimulation` then combines the exact `currentAmounts` strings through `Number(amountA)` / `Number(amountB)` and `makeDisplayTokenAmount`, which also returns a number (around lines 110–150).

With the reported value, `Number("62667447936.264477").toString()` is `"62667447936.26448"`, already 3 raw units high for a 6-decimal token. Converting the later subtraction result with `inputAmount.toFixed()` cannot recover those digits, so the reposition route/broadcast path is not actually string-precise end-to-end yet.

Please keep the amount outputs and arithmetic in these reposition helpers as decimal strings/BigNumber values, use the string display conversion for raw amounts, and add a regression that observes the exact amount passed from a large reposition balance into its route or swap request.

jinoosss:
Addressed in the latest commit: the reposition helpers are now string/BigNumber end-to-end.

**`reposition-utils.ts`**
- `getRepositionAmountsByPriceRange` returns `{ amountA: string; amountB: string }`. The deposit-ratio arithmetic is done in BigNumber (no `Number(...)` / `.toNumber()`); the 6-decimal cases use `decimalPlaces(6).toFixed()` and the default case returns the exact product. `Infinity` cases return the string `"Infinity"` (still handled by `BigNumber(...).isFinite()` downstream).
- `getRepositionAmountsWithSwapSimulation` takes and returns string amounts, adds the swap output via `makeDisplayTokenAmountString` + BigNumber, and converts deposit results with `BigNumber(raw).shiftedBy(-decimals).toFixed()` instead of `makeShiftAmount`.
- `toShiftBitInt` rounds in BigNumber (`integerValue(ROUND_HALF_UP)`) instead of `Math.round` on a JS number.
- The route-estimate request and the swap request amounts are extracted into pure helpers `makeRepositionSwapEstimateRequest` and `makeRepositionSwapAmounts`, which `use-reposition-handle` now calls. This is the production code path for both requests.

**Regression coverage (`reposition-utils.spec.ts`, using `62667447936.264477` / raw `62667447936264477`)**
- `getRepositionAmountsByPriceRange` keeps the large balance exact when the range is unchanged (previously `.toNumber()` gave `62667447936.26448`).
- The route request carries the exact remainder: `0.000001` -> raw `1` when the target is one raw unit lower, and the full `62667447936.264477` -> raw `62667447936264477` when everything is swapped.
- The swap request carries the exact large `tokenAmount` and string limits (`"0.995"` for EXACT_IN; the EXACT_OUT max input rounded up).
- The swap simulation adds the swap output to the large balance without rounding.

Full suite: 168 suites / 767 tests passing; no new tsc or eslint errors.

junghoon-vans (approved):
Verified the latest update. The reposition path now keeps token amounts as decimal strings/BigNumber values through range calculation, swap simulation, route-request construction, and broadcast-request construction. `toShiftBitInt` also performs the atomic-unit conversion without a JavaScript-number round trip.

The new regression coverage exercises the reported large balance through unchanged-range calculation, a one-raw-unit remainder, a full-balance route request, exact-in/exact-out limits, and simulated output addition. I also ran the affected precision suites locally: 5 suites and 63 tests passed.

The requested precision corrections are complete from my side.

PR #989, opened 2026-09-15
refactor(governance): split CollectReward into per-token protocol fee reward messages

jinoosss:
## Summary

- Replace the single `gov/staker.CollectReward` call with per-token multi-messages:
  - `gov/staker.CollectProtocolFeeReward(tokenPath)` for each `claimableGovernanceRewards` entry with `amount > 0`
  - `launchpad.CollectProtocolFeeReward(tokenPath)` for each `claimableLaunchpadRewards` entry with `amount > 0`
- Remove the launchpad `CollectProtocolFee` message from the governance claim flow
- Add message builders for `gov/staker.CollectEmissionReward` and `launchpad.CollectEmissionReward` (not wired yet; emission claiming will be added separately)
- Change `sendCollectReward` to accept the claimable reward arrays from the API response instead of two booleans, and update the hook / component / mock accordingly

## Test plan

- [x] `jest src/repositories/governance src/repositories/launchpad` passes (including zero-amount filtering)
- [x] No new tsc or lint errors in changed files
- [ ] Manually verify claiming on the governance page emits one message per claimable token


packages/web/src/repositories/governance/governance-repository-impl.ts
```diff
@@ -433,13 +439,19 @@ export class GovernanceRepositoryImpl implements GovernanceRepository {
...
   ): Promise<WalletResponse<{ hash: string }>> => {
     const caller = await this.getAddress();
     const messages = [
-      ...(claimGovernanceRewards ? makeCollectRewardMessages({ caller }) : []),
-      ...(claimLaunchpadRewards ? makeCollectProtocolFeeMessage({ caller }) : []),
+      ...makeCollectProtocolFeeRewardMessages({
+        tokenPaths: getClaimableTokenPaths(claimableGovernanceRewards),
+        caller,
+      }),
+      ...makeCollectProtocolFeeRewardFromLaunchPadMessages({
```
junghoon-vans: This direct call to `gov/staker.CollectProtocolFeeRewardFromLaunchPad` will fail its authorization check. The contract requires `rlm.Previous().Address()` to be the launchpad realm (`access.AssertIsLaunchpad`), but in this transaction the previous realm is the user.

Please route each launchpad reward through the public launchpad entrypoint instead:

`launchpad.CollectProtocolFeeReward(tokenPath) -> gov/staker.CollectProtocolFeeRewardFromLaunchPad(caller, tokenPath)`

The repository test currently only asserts the generated message shape, so it encodes this unauthorized route without exercising the contract guard.

junghoon-vans (approved):
Launchpad collection routing needs to be corrected as noted in the inline comment. Emission reward collection will be handled in a follow-up.

notJoon (approved):
not blocker: `makeCollectProtocolFeeMessage` and `makeLaunchpadCollectEmissionRewardMessages` function seems dead code now. so consider to remove them.

PR #996, opened 2026-09-18
feat(rpc): fail over to an optional fallback RPC when the primary is down

jinoosss:
## Summary

Adds an optional `NEXT_PUBLIC_DEFAULT_CHAIN_FALLBACK_RPC_URL`. When a request against the primary `rpcUrl` fails at the transport level, the provider retries on the fallback and keeps using it; if the fallback also fails, it rotates back to `rpcUrl`.

```dotenv
NEXT_PUBLIC_DEFAULT_CHAIN_RPC_URL="https://dev.rpc.gnoswap.io"
# Optional. Used only when the primary RPC stops answering.
NEXT_PUBLIC_DEFAULT_CHAIN_FALLBACK_RPC_URL=""
```

The variable is optional: left unset, `NetworkModel.fallbackRpcUrl` stays `undefined` and every path below behaves exactly as it does today, against a single endpoint.

This ports [adena-wallet#907](https://github.com/onbloc/adena-wallet/pull/907) to this codebase.

## Failover behaviour

`RpcEndpointSelector` (new, `common/clients/gno-provider/rpc-endpoint-selector.ts`) treats a network's endpoints as a ring:

1. Requests start on `rpcUrl`.
2. A transport failure rotates to `fallbackRpcUrl` and retries the same request there. The rotation is **sticky**, so the retry and every later request go straight to the fallback instead of paying the dead endpoint's timeout again.
3. A transport failure on `fallbackRpcUrl` rotates back to `rpcUrl` in the same way.
4. Each `run` tries every endpoint at most once, so a network that is down everywhere still fails fast rather than looping.

Only **transport** failures rotate — an unreachable host, a 5xx/429 response, a timeout. Errors the chain actually answered with (insufficient funds, a rejected tx, a 4xx on a malformed query) are rethrown untouched: rotating on those would move off the primary endpoint over a failure the fallback is going to reproduce anyway.

## Coverage

`FallbackRpcClient` (new) is an `RpcClient` holding one `HttpClient` per endpoint and routing through the selector. It is what the `Tm2Client` is built on, so the JSON-RPC calls the base `GnoJSONRPCProvider` issues — `getAccount`, `evaluateExpression`, the `abciQuery` behind `getGasPrice` — fail over too.

Two gaps this codebase has that the endpoint ring alone does not cover:

- tm2-rpc's `HttpClient` passes no `AbortSignal`, so a blackholed node hangs forever and the selector never gets the failure it needs to rotate. Each attempt now carries its own deadline (default 10s), and the timeout it raises is classified as a transport failure. The request itself keeps running until the browser drops it; what this guarantees is that the caller always settles.
- `Tm2Client.create(rpcClient)` is just `new Tm2Client(rpcClient)` — unlike `Tm2Client.connect(url)`, which this provider used before, it never touches the node. Swapping one for the other would have handed out a provider for an endpoint that is not answering at all, and the `RpcConnectionFailed` screen would never show. A `status()` query after construction keeps that round trip, and since it runs through the failover a dead primary endpoint is left behind at connect time rather than at the first query the app makes.

Transaction result polling (`TransactionEventStore`) goes through the selector as well, so a node going down mid-transaction does not leave the receipt pending forever. Its constructor now takes the slice of the client it actually uses (`TransactionQueryClient`) instead of an `Axios`; an Axios instance still satisfies it, so the existing spec is unchanged.

## Not ported

The adena PR's `chains.json` migration and `fetchHealth` changes have no counterpart here: networks come from env vars and a bundled constant rather than local storage, so there is no stored entry to graft a field onto, and there is no health-check banner.

## Tests

- `rpc-endpoint-selector.spec.ts` (new) — rotation, stickiness, rotating back after the fallback fails, 5xx vs 4xx classification, chain-level errors not rotating, no-fallback and duplicate-fallback cases.
- `fallback-rpc-client.spec.ts` (new) — retrying on the fallback endpoint, one HTTP client per endpoint, and rotating away from a node that never answers.

`tsc --noEmit` and `eslint` report no new issues; the full suite passes (173 suites / 799 tests).

## Summary by CodeRabbit

* **New Features**
  * Added optional fallback RPC endpoint configuration for improved network reliability.
  * Automatically switches to the fallback endpoint when the primary endpoint is unavailable or unresponsive.
  * RPC requests now use timeout protection and can recover when the primary endpoint becomes available again.
  * Transaction-related network requests also support primary and fallback RPC endpoints.

* **Tests**
  * Added coverage for endpoint failover, retries, timeouts, recovery, and error handling.


junghoon-vans (changes requested):
Requesting changes. I confirmed four blocking failover gaps already called out inline: (1) per-attempt Promise.race timeouts do not abort the underlying tm2-rpc fetch, so repeated failures can accumulate pending requests; (2) the default 15s connection budget truncates the second 10s endpoint attempt to 5s; (3) generic message matching can classify a JSON-RPC application error containing “timeout” as a transport failure and retry it; and (4) transaction polling uses axios.get with the default infinite timeout, so a blackholed primary never rejects and never reaches the fallback. The focused selector/client suites pass (20 tests), but these production failure modes remain unhandled.

jinoosss:
Addressed in 3bcae0830. Four of the five findings held up; one did not.

**Abort the underlying request (`fallback-rpc-client.ts`)** — fixed by dropping tm2-rpc's `HttpClient` as the transport. It passes no `AbortSignal` and its `disconnect` is a no-op, so nothing short of replacing it actually cancels the request; `Promise.race` only rejected the caller. Requests now go out over axios with `timeout`, which aborts the request itself. This also removes the hand-rolled deadline and the per-endpoint client map.

**Classify by shape, not by message (`rpc-endpoint-selector.ts`)** — fixed, and the message matching is gone entirely rather than narrowed. A node answering with a JSON-RPC error now surfaces as `JsonRpcResponseError` and is rethrown untouched; only an axios error with no response (refused connection, DNS failure, aborted timeout) or a 5xx/429 rotates. Anything else no longer fails over at all, which is a deliberate tightening: every transport this selector wraps is now axios.

**Connection budget (`gno-provider.ts`)** — fixed along the lines suggested, with the endpoint count read off the selector rather than recomputed, so the de-duplication of a fallback equal to the primary stays in one place:

```ts
const connectTimeoutMs =
  options.connectTimeoutMs ?? Math.max(MIN_CONNECT_TIMEOUT_MS, requestTimeoutMs * endpoints.count);
```

**Transaction polling deadline (`GnoswapServiceProvider.tsx`)** — fixed, using the same `RPC_REQUEST_TIMEOUT_MS` the provider's requests use.

**Request-local endpoint sequence (`rpc-endpoint-selector.ts`)** — the interleaving described is not reachable. `rotateFrom` and the next read of `activeIndex` are in the same synchronous block: the `catch` runs on rejection, the loop condition and the index read follow with no `await` in between, so no other run can be scheduled there. I could not write a test that reproduced it, which is what surfaced this.

I took the change anyway, as it makes "each run walks every endpoint at most once" hold on its own rather than resting on that argument about suspension points, but there is no fixed bug behind it and no test claiming one.

Tests updated for the new transport; the suite passes (173 suites / 800 tests), `tsc --noEmit` and `eslint` are clean.

junghoon-vans (approved):
Verified the follow-up changes at 3bcae083. All four previously identified failover issues are addressed: requests now use abortable Axios timeouts, the connection budget covers every endpoint attempt, JSON-RPC application errors are distinguished from transport failures, and transaction polling has a finite timeout. The focused RPC suites pass (2 suites, 21 tests), and I found no new blocking issue in the follow-up diff.

PR #1000, opened 2026-09-21
chore: change default value of withReset in makeTransactionMessagesWithApproves to false

jinoosss:
## Descriptions

Remove the message that resets the GRC20 Approve for each transaction.

GnoSwap Interface do not grant Approve to any contracts other than the gnoswap contract, and we only grant Approve for the necessary amount. (We do not grant Approve for excessive amounts, such as MaxInt64.)

On Ledger devices, the transaction may not execute properly due to long messages caused by the JSON token count and message view limits.

## Summary by CodeRabbit

- **Bug Fixes**
  - Transaction approval flows now include only approval and transaction messages by default.
  - Reset approval messages are included only when explicitly enabled.


junghoon-vans (approved):
Approved after review.

